PR #11, opened 2012-06-20
Hi! I think there was a bug on Router.php :)

javierhonduco:
The old regex, used to route  /love and /love/ when was defined Flight::route("/love", function(){})

### DIFF
--- a/flight/net/Router.php
+++ b/flight/net/Router.php
@@ -80,7 +80,7 @@ class Router {
         $ids = array();
 
         // Build the regex for matching
-        $regex = '/^'.implode('\/', array_map(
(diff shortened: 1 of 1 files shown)


mikecao:
This is by design, the urls /love and /love/ should return the same content. Github does the same thing. If you add a / to the url up top it will show the same content.

Additionally, your fix isn't correct. You're removing the beginning of the regular expression, /^. It works, but I think it's just a side effect. If you really want ending slashes to be treated uniquely you can look for this code in Router.php:

)).'\/?(?:\?.*)?$/i';

and change it to:

)).'(?:\?.*)?$/i';

Removing the "\/?" means the ending slash is not optional.

PR #15, opened 2012-10-22
Add composer support

davidkuridza:
Added support for http://getcomposer.org/.


mikecao:
Shouldn't the line that says `flight/flight.php` actually be `flight/Flight.php`?

davidkuridza:
You are correct, was testing something and overlooked the change when pushed it. Feel free to squash both commits into one :)

PR #20, opened 2012-12-20
Fixed muted error in Loader

ilsenem:
This error appear only when default error handler is rewriten or custom error handler assigned.

Despite that there is no error in normal occasion, IMHO have any error in code is bad even if you don't see one.

Also it is strange to ignore some types of errors in error handler. Maybe configuration could regulate it - `flight.debug` or `flight.production` for example?

P.S. Sorry for my english.

### DIFF
--- a/flight/core/Loader.php
+++ b/flight/core/Loader.php
@@ -175,7 +175,8 @@ class Loader {
         }
 
         // Allow other autoloaders to run before raising an error
-        $loader = array_pop(spl_autoload_functions());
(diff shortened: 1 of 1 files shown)


mikecao:
You're right, the error handler shouldn't be ignoring certain types of errors. I've changed it to accept any errors set by the error_reporting() function. I've also fixed some additional E_NOTICE and E_STRICT errors, including the one you found. Thanks.

PR #22, opened 2013-01-09
a few changes to request.php

kafene:
I wrote a big summary of them and now I have an empty form looking at me for this pull request. Here's a direct link to the changes:

https://github.com/kafene/flight/commit/5270c6b23a1fb62e4b44523f99bc87255fed1d45


kafene:
ah, that is right. I saw your note on the line and went back in to test it, and indeed it should be returning instead of setting it. I will make the correction. Thank you.

PR #25, opened 2013-01-22
Update flight/template/View.php

ghost:
so do be aware that this template file


warmwaffles:
This isn't really necessary... why not just include `myfile.html` it becomes more explicit in code as to what you are doing.

mikecao:
Shouldn't force people to name their files `file.html.php`, since it will may not always be html being rendered.

warmwaffles:
Troll face for his avatar, very appropriate :D

PR #33, opened 2013-04-09
Removing redundant code that repeats actions for different number of parameters passed

aramboyajyan:
On several places the `Dispatcher` and `Loader` methods repeat the same action for different number of parameters passed. Finally, default handler works for any number of parameters.

This code seems to be redundant as the framework works as expected without those lines.

Have a look please and merge if works for you as well.

Thanks.

### DIFF
--- a/flight/core/Dispatcher.php
+++ b/flight/core/Dispatcher.php
@@ -152,22 +152,7 @@ class Dispatcher {
      * @return mixed Function results
      */
     public static function callFunction($func, array &$params = array()) {
-        switch (count($params)) {
-            case 0:
-                return $func();
-            case 1:
-                return $func($params[0]);
-            case 2:
-                return $func($params[0], $params[1]);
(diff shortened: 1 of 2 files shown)


mikecao:
The code is like that for performance. Calling call_user_func_array is much slower than invoking the function directly, on average 3 times slower. The same is true with invoking ReflectionClass. It checks up to 5 parameters before trying the defaults, which should cover a majority of cases.

warmwaffles:
I agree with @mikecao. I've seen this done in other projects other than PHP for the exact same reason. It does obfuscate a bit, but the performance gain is worth it.

warmwaffles:
I agree with @mikecao. I've seen this done in other projects other than PHP for the exact same reason. It does obfuscate a bit, but the performance gain is worth it.

aramboyajyan:
Thanks for clarification.

PR #42, opened 2013-06-25
Defined 'flght' namespace

ghost:


mikecao:
The latest version now includes a separate namespaced Engine class, while leaving the original Flight class a static global.

PR #46, opened 2013-08-06
PSR-0 Compatible and removed autoloading php file in composer.json

slashmili:
n order to use psr-0 in composer we need all this changes!

BTW after this changes for running unit test you need to run

```
$ composer.phar --dev install
$ ./vendor/bin/phpunit tests
```

FYI we added Flight to our composer and our app didn't work properly any more! because of this composer setting:

```
"autoload": {
   "files": ["flight/Flight.php"]
}
```

This way composer forcefully load Flight into our all PHP codes, even though we didn't use Flight on that particular PHP.


slashmili:
I suppose I forgot to commit all the files  !

PR #47, opened 2013-08-06
Fix composer autoloader

boynux:
This patch fixes composer autoloader issue, which is calling Flight::init() before any try to access Flight classes.
This patch postpones init() call to first Flight class access. (does not affect normal Flight.php inclusion)


mikecao:
The latest version now no longer calls init() automatically.

PR #49, opened 2013-08-07
Support for Object cloning

shushant:
To reduce the cost of instantiating objects we can simply clone the parent object


mikecao:
I'm not convinced that cloning produces a significant performance improvement over creating a new object. One calls __construct() while the other runs __clone(), but they both simply create new objects.

PR #62, opened 2013-10-29
JSONP support

ozh:
Hi

Thanks a zillion for this neat little framework. I'm using it to develop api.yourls.org, it's lovely and has everything needed to roll out APIs :-)

I would like it to support JSONP out of the box, would you be interested in a pull request?

My implementation works this way:
- new default setting to set the callback name: `flight.jsonp.callback`
- new function `_jsonp()` outputting JSON padded with the callback value (eg `/route/?jsonp=foo` will output `foo(...JSON response here...);`

Thanks for your time!

### DIFF
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -447,7 +447,7 @@ class Engine {
      */
     public function _jsonp($data) {
         // Get the callback value (eg '?jsonp=my_function') and pad the output
-        $callback = \Flight::request()->query[ $this->get('flight.jsonp.callback') ];
+        $callback = $this->request()->query[ $this->get('flight.jsonp.callback') ];
         $this->response()
             ->status(200)
             ->header('Content-Type', 'application/javascript')
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -94,7 +94,7 @@ class Engine {
         // Register framework methods
         $methods = array(
             'start','stop','route','halt','error','notFound',
-            'render','redirect','etag','lastModified','json'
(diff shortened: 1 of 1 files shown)


ozh:
thanks!

ozh:
The doc needs updating now, in http://flightphp.com/learn#configuration and http://flightphp.com/learn#methods

kamov:
Thanks!

An example of usage will be useful.

Thanks!

ozh:
It works the same as the `json` method. Something like:

``` php
// Set the JSONP callback parameter to "callback" (default: "jsonp")
Flight::set( 'flight.jsonp.callback', 'callback' );

Flight::route('/myroute', function(){
    $data = function_to_get_some_data();
    Flight::jsonp( $data );
});
```

And then request url `server.com/myroute?callback=my_function` to get a JSONP response

A more practical example would be to check for the callback parameter and deciding whether to output json on jsonp. Full example:

``` php
<?php
require 'flight/Flight.php';

// Set the JSONP callback parameter to "callback" (default: "jsonp")
$callback = 'callback';
Flight::set( 'flight.jsonp.callback', $callback );

Flight::route('/myroute', function(){
    $data = function_to_get_some_data();
    if( Flight::request()->query[ $callback ] ) {
        Flight::jsonp( $data );
    } else {
        Flight::json( $data );
    }
});

Flight::start();
?>
```

kamov:
@ozh thanks! This help!

PR #70, opened 2013-12-17
Use $_SERVER instead of getenv.

eric1234:
See commit message for full explanation.

Your automated tests still pass after this patch is applied and I have done limited testing with my app to verify it doesn't break anything and solves the problem I am trying to solve. But since it was just written it has not gotten extensive testing.

If you want I can ping this issue when my app is done which will provide more confidence that nothing is broken by this patch.

### DIFF
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -466,7 +466,8 @@ class Engine {
 
         $this->response()->header('ETag', $id);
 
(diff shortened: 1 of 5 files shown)


ozh:
:+1: 

Quick research on `getenv()` seems to show lots of problems (with IIS, PHP CLI to name a few)

PR #87, opened 2014-01-10
Simplify output buffering code

wpscholar:
No need to use both ob_get_contents() and ob_end_clean() when you can just use ob_get_clean().


mikecao:
Thanks

PR #111, opened 2014-04-07
Added $params property to Request with merged GET, POST, PUT and DELETE ...

darquill:
...params.

I thought it would be useful to have a single Collection with all the GET, POST and raw (PUT and DELETE from php://input) parameters merged together. This would allow for a layer of abstraction between the method of a request and the mapped function.

### DIFF
--- a/flight/net/Request.php
+++ b/flight/net/Request.php
@@ -109,6 +109,11 @@ class Request {
     public $files;
 
     /**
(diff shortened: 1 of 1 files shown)


cceconi:
I think this feature can be easily done outside the framework (request extension) and its better because flight don't need to assume name collision.

darquill:
Closing the pull request since dev seems uninterested and @cceconi makes a good point.

PR #116, opened 2014-05-23
Add support for parsed JSON property in Request.

berkus:
Make it really handy to use Flight in RESTful APIs with JSON payload.

Just access Flight::request()->json.


berkus:
Updated the patch.

PR #121, opened 2014-06-26
Fixed problem with Request Json array

liamjack:
Using the Flight Framework with AngularJS, I send POST data as JSON in the request body, my problem was that the browser sent the following content type : "application/json;charset=utf-8".

When this was interpreted by the Request handler, it didn't get put into the Flight::request()->json variable because the content type didn't exactly match " application/json". I have fixed this and now also send the json data through the Collection system to enable users to access as an object or an array, as they desire.

### DIFF
--- a/flight/net/Request.php
+++ b/flight/net/Request.php
@@ -187,8 +187,11 @@ class Request {
 
         if (strpos($this->type, 'application/json') === 0 && $this->body != '') {
             $this->json = json_decode($this->body, true);
-			$this->json = new Collection($this->json);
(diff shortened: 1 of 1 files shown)


liamjack:
Sorry about that, having problems with my client. Updated now :)

kafene:
Might I suggest

```
if (strstr($this->type, ';', true) == 'application/json'
```

I get a ton of Google results for "application/jsonp" which would match testing for strpos == 0. Although its unlikely to be the request type.

PR #124, opened 2014-07-04
added new setting "add_route_obj"

smeinecke:
To be able to use the same callback multiple times (with default values for parameters) like this:

```
Flight::route('GET /', array('\Application\Controller\Section', 'showSection'));
Flight::route('GET /section/@name', array('\Application\Controller\Section', 'showSection'));
Flight::route('GET /section/@name/@subname', array('\Application\Controller\Section', 'showSection'));
```

It would be nice to restore the old behavior of flightphp to disable auto append of Route object as last parameter to every route callback. 

By setting 'flight.add_route_obj' to false this is possible with this patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -795,6 +795,7 @@ The following is a list of all the available configuration settings:
     flight.handle_errors - Allow Flight to handle all errors internally. (default: true)
     flight.log_errors - Log errors to the web server's error log file. (default: false)
     flight.views.path - Directory containing view template files. (default: ./views)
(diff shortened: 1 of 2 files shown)


mikecao:
I've implemented similar functionality in commit d84e51ba4739a31f67755935124bd373dc43040c. Instead of a global config, you can do it at the route level by passing in `true` as a third parameter:

``` php
Flight::route('/', $callback, true);
```

smeinecke:
Thank you for your work - this solves my problem as well :+1:

PR #125, opened 2014-07-04
added workaround to load Request body only if needed 

smeinecke:
On every call routed by flightphp, the Request body will be read by the Request object (even if it is not used at all). This is very time (CPU cycles) and memory consuming if you use file upload or PUT requests as the whole data get loaded into memory in one chunk.

Additional it is possible to crash the whole php application by only sending enough data as the request could not be parsed anymore.

Added workaround to load body only if it's needed (and not loaded by the application itself).

### DIFF
--- a/flight/net/Request.php
+++ b/flight/net/Request.php
@@ -77,7 +77,7 @@ class Request {
     /**
      * @var mixed Raw data from the request body
      */
-    public $body;
+    protected $_body;
 
     /**
      * @var string Content type
@@ -130,6 +130,11 @@ class Request {
     public $proxy_ip;
 
     /**
+     * @var bool true if body already load
+     */
+    protected $_body_loaded = false;
+
+    /**
      * Constructor.
      *
      * @param array $config Request configuration
@@ -146,7 +151,6 @@ class Request {
                 'ajax' => self::getVar('HTTP_X_REQUESTED_WITH') == 'XMLHttpRequest',
                 'scheme' => self::getVar('SERVER_PROTOCOL', 'HTTP/1.1'),
                 'user_agent' => self::getVar('HTTP_USER_AGENT'),
-                'body' => file_get_contents('php://input'),
                 'type' => self::getVar('CONTENT_TYPE'),
                 'length' => self::getVar('CONTENT_LENGTH', 0),
                 'query' => new Collection($_GET),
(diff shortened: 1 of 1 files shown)


mikecao:
I went with a different approach with commit ab2aba4e7dca554444003d49569dce9f7bbda88b. It will only load the body on requests that will potentially have a body (POST and PUT). File uploads are not an issue because PHP does not populate the body for multipart/form-data.

smeinecke:
You probably misunderstood my request. If I send a PUT request with 25 MB as payload (body) to any routed URL, your fix will load this in one big chunk into memory and I can not do anything about this as this is done before even any routing is done.

If you have to handle PUT requests in PHP, you want to read the data in chunks of 4 kByte to not load the whole data into one single variable into memory :-)

mikecao:
Your way is slightly more efficient in that it doesn't load the body until you need it, but it's doing it the exact same way using `file_get_contents` so you'll run into the same problem. I think if people are really concerned they can already control the max request size via php.ini.

smeinecke:
I only implemented it that way as I did not want to break any code thats already using the body in Request. Normally you would like to do something like:

``` php
Flight::route('PUT /my/route', function() {
    $hIn = fopen('php://input', 'rb');
    $hOut = fopen('my_output/file.bin', 'wb');
    while (!feof($hIn)) {
        $data = fread($hIn, 8192);
        fwrite($hOut, $data);
    }
    fclose($hIn);
    fclose($hOut);
});
```

Or process the data while reading it as stream, but not holding the full content in memory.
Also, if I limit the max request size in PHP - I can not control the size for only one URL in my whole project. 

So, if I have some subpage in my project, where I need to upload more than 2 MB of data by PUT (which is a standard situation for a REST API), it would be still possible to exploit the whole server by sending 50MB as request body by 30 connections parallel, as the whole 1.5 GB would be loaded directly into memory. It is also possible, that the whole stuff would be loaded in duplicate as I do not know in detail how PHP internally handle that stuff.

I am sorry that I have to insist upon this, as I think this is a real security issue I do not want to leave open in any of my projects :-)

mikecao:
The request body would have already been processed by the webserver before it gets to PHP, so it's already sitting in memory. It won't prevent someone from sending you a large payload using multiple connections, basically a DDOS. I guess worse case you would duplicate the amount in memory on the PHP side. 

But I can understand your point about wanting to process the request body your own way. So I think the solution is to remove the property and simply have people call `getBody()` or process it themselves.

smeinecke:
If you can do that, that would be much better and we do not net some ugly workaround like done in my code :+1:

smeinecke:
Thank you for your fix in da40e03eb4a39745107912dffe926a8fce0d38dc - but now I can trigger the same issue by simply sending a JSON content type in PUT/POST Request.

nginx and php-fpm for example does not read the full request into memory if you send an PUT request. You get the body payload as stream of data and only if you read from STDIN, if not, it will be routed to /dev/null.

mikecao:
There needs to be some kind of balance. You can't deny functionality because somebody may not have an optimal setup. You can already control the payload size with the setting `client_max_body_size` in nginx and `post_max_size` in PHP. What if the user allows unlimited POST size on their server, do we never process $_POST? I think if the user is explicitly sending a JSON request we need to process it. For any other request, it is up to you how to handle it.

smeinecke:
Okay - we will not agree in this any more. I close this pull now and have to maintain my own fork of flight.

PR #138, opened 2014-08-17
Extra parameters in Route

kouts:
Added the ability to pass extra parameters in Route. 
Use it like this:
Flight::route('/', function(){
    echo 'hello world!';
}, false, array('lang'=>'en', 'directory'=>'backend'));

### DIFF
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -392,8 +392,8 @@ class Engine {
      * @param callback $callback Callback function
      * @param boolean $pass_route Pass the matching route object to the callback
      */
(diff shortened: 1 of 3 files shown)


mikecao:
If you need to pass data, you should just use the set method:

``` php
Flight::set('data', array('lang'=>'en', 'directory'=>'backend'));

Flight::route('/', function(){
  $data = Flight::get('data');
});
```

kouts:
I'm trying to pass the extra parameters to controllers not closures, can you give an example of how to to this?

PR #141, opened 2014-08-20
replacing array_push with faster []

pine3ree:


ozh:
This notation requires 5.4+

philo23:
I don't think so, this syntax has been in PHP for a long time.

You might be thinking of the new array syntax that replaces using `array()`

pine3ree:
it's a simple standard array element assignment. It has nothing to do with php 5.4 or new array notation. [] array operator is as twice as fast as array_push.

ozh:
My bad, [indeed](http://3v4l.org/nfGCE).

pine3ree:
@ozh 

no problem. :-)

PR #143, opened 2014-09-05
Fixed splat

kouts:
Fix for Route splat when a named param is present in a route pattern.


mikecao:
I decided to use a different implementation, but thanks for reporting it. See commit 88f4c1fb7108d11240e2db341be641f388d52124.

kouts:
Great, thanks!

PR #151, opened 2014-09-30
PUT only allows php://input to be read once

dstelljes:
On PUT requests, the contents of the input stream can only be read once. That means `Request::getBody()` can't be called multiple times. Is this a reasonable workaround?

### DIFF
--- a/flight/net/Request.php
+++ b/flight/net/Request.php
@@ -121,11 +121,6 @@ class Request {
     public $proxy_ip;
 
     /**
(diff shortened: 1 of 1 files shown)


mikecao:
I would make the `$body` variable static inside the method. That way you can use and return the same variable instead of having to copy it back to `self::$body`.

dstelljes:
Thanks for the feedback. How's this?

PR #156, opened 2014-10-30
Added list of methods for autocomplete

terales:
Added list of available methods to the PHPDoc section of `Flight` class. Inspired by suggestion: http://stackoverflow.com/a/26628288/1363799

Autocomplete was tested in PHPStorm 8.0.1:

flight-autocomplete

### DIFF
--- a/flight/Flight.php
+++ b/flight/Flight.php
@@ -8,6 +8,28 @@
 
 /**
  * The Flight class is a static representation of the framework.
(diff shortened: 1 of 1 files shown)


mikecao:
Nice work

ghost:
Could we get a tag on this for packagist? I'd prefer not to use dev-master in my projects.

PR #160, opened 2014-11-17
Add previousExceptionHandler attribute in Engine class

cceconi:
Add previousExceptionHandler attribute in Engine class to allow integration with third party lib
ex : monolog

Very usefull for me.


cceconi:
Hi,
I would like to know if you will integrate this pull request ? Because I must maintain my fork currently just for that.
Thank you and good job.

cceconi:
This code will NOT always run the previous exception handler but only if one has been set previously.

```
if ($this->previousExceptionHandler) {
    call_user_func($this->previousExceptionHandler, $e);
}
```

Alternatively, using `Flight::before('error', ...)` will not solve this need and `restore_exception_handler` too drastic since your implementation is generic and already extensible.
The objective is to be more agnostic and not opinionated, so I get your point, developer can choose.

## First option

In the init function :  

```
// Default configuration settings
$this->set('flight.allow_previous_exception', false);
```

In the handleException function : 

```
if ($this->get('flight.allow_previous_exception') && $this->previousExceptionHandler) {
    call_user_func($this->previousExceptionHandler, $e);
}
```

## Second option

In the handleException function :

```
$this->error($e, $this->previousExceptionHandler);
```

And change your _error signature : 

```
public function _error(\Exception $e, $previousExceptionHandler = null)
```

The second option is probably more flexible but requires that developer overide the error function to manage it.
For my part, I prefer the second one because the responsibility of execute the previousExceptionHandler function is not in Flight and come from a specific need.

mikecao:
I don't understand why `Flight::before('error', ...)` wouldn't work for you. You're getting passed a copy of the exception and you can do whatever you want with it, including passing it any other exception handlers. Your code seems to do that same thing.

cceconi:
The problem is that the exception handler is not defined by me but somewhere in one or some third party libs. So it's easier  and more secure to call it having the $previousExceptionHandler function than finding it and initialize it correctly.

mikecao:
I still think the before error method is the better option since it is explicit code and you know exactly what it will do. If you are using multiple third party libraries and they all use `set_exception_handler` then you will not know what the `$previousExceptionHandler` refers to. But if you use the hook you can do this:

``` php
Flight::before('error', function($err){
  $library1->handleError($err);
  $library2->logError($err);
});
```

cceconi:
I'll give a try and a feedback as soon as possible.

PR #172, opened 2015-02-10
Added phpdoc for the response method

ghost:
Title should speak for itself


tamtamchik:
@mikecao @TheSerenin implemented in #226

PR #178, opened 2015-03-23
include pragma='no-cache' in the cache header reset

sethbattin:
Once `\flight\Response::cache` has been called with argument `$expires = false`, it cannot be reset because of the header `"Pragma=no-cache"`.  This change removes that header when `cache` is called with a non-false `$expires`.

Example case: selectively enabling cache for ajax requests.  In `\flight\Engine::_start`, all ajax requests disable the cache (see comment `// Disable caching for AJAX requests`).  Re-enabling cache for a given request is not possible (cleanly), because the pragma header persists.  After this change, an ajax request can re-enable http cache by calling `\flight\Response::cache` with a future time value.

### DIFF
--- a/flight/net/Response.php
+++ b/flight/net/Response.php
@@ -190,6 +190,9 @@ class Response {
             $expires = is_int($expires) ? $expires : strtotime($expires);
             $this->headers['Expires'] = gmdate('D, d M Y H:i:s', $expires) . ' GMT';
             $this->headers['Cache-Control'] = 'max-age='.($expires - time());
+            if (isset($this->headers['Pragma']) && $this->headers['Pragma'] == 'no-cache'){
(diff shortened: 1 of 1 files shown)


kafene:
Better yet, just remove the `Pragma` header since it's useless in HTTP 1.1; `Cache-Control` takes precedence.

kafene:
I think the better course of action here is to stop sending the `Pragma` header altogether. Since HTTP 1.1, `Cache-Control` should be used instead.

sethbattin:
That would be fine for a feature change.  It's unlikely to have any breaking changes, but it still ought to have some thorough tests to make certain.  

I still would recommend my change as an immediate bugfix, regardless.  The framework is currently broken for any browser that respects `Pragma` in a response, which includes Chrome 41.

PR #196, opened 2015-06-29
Anti-XSRF token property added to Request object

mivanov93:
Added an Anti-XSRF token property to Request object which we get from the HTTP headers, namely - X-XSRF-TOKEN. We need it to verify that our request is not from an auto-submitted form.


Conver:
Not everyone has the HTTP_X_XSRF_TOKEN

PR #197, opened 2015-07-17
delete this pull request, made it from the wrong branch

s4wny:
Delete this pull request, made it from the wrong branch. The correct pull request can be found at https://github.com/mikecao/flight/pull/198

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "mikecao/flight",
+    "name": "s4wny/flight",
     "description": "Flight is a fast, simple, extensible framework for PHP. Flight enables you to quickly and easily build RESTful web applications.",
     "homepage": "http://flightphp.com",
     "license": "MIT",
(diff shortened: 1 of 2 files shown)


ymakux:
Maybe specify base directory in RewriteBase? My localhost installation works fine in subfolder out of  box

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "mikecao/flight",
+    "name": "s4wny/flight",
```
tamtamchik, line 2: wow)

s4wny, line 2: lol, apparently github automatically adds all commits on the master branch to this pull request.

Added that line so I can install my modifed version via composer. Lesson learned: always use a feature branch for pull requests.

s4wny:
Making a new pull request from a feature branch instead. Of course I do not want to change the composer.json file in my pull request.

The correct pull request can be found at https://github.com/mikecao/flight/pull/198

PR #198, opened 2015-07-18
Update Request.php to make flightphp working even if your documentroot is a subfolder

s4wny:
# Case

I want to set the root directory to the subdirectory `/public` so users can't access config files etc. This didn't work for me with the current FlightPHP code, so I made a path to make it work for my case. I think this change won't break old code but I'm not sure. I'm pusing it anyway so it show up on google for everyone who have this problem.
## My setup

```
./my-project
    /public
        /css
        /js
        .htaccess         [1]
        app.php           // FlightPHP app is here
        ...
    /vendor
    /config
    .htaccess             [2]
    ...
```

.htaccess [1]

```
RewriteEngine On
RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{REQUEST_FILENAME} !-d
RewriteRule ^(.*)$ app.php?/$1 [L]
```

.htacces [2]

```
RewriteEngine On

RewriteCond %{REQUEST_URI} !public/.*$
RewriteRule ^(.*)$ public/$1 [L]
```

So when a user go to `http://website.com/my-project/` apache will actually load `http://website.com/my-project/public` which will load `http://website.com/my-project/app.php`.

`http://website.com/my-project/admin/action/xd` apache will actually load `http://website.com/my-project/public` which will load `http://website.com/my-project/app.php/admin/action/xd`.

`QUERY_STRING` = `/admin/action/xd/`.

Note that you still need to set `Flight::set('flight.base_url', '/my-project/');` to make redirects work.

### DIFF
--- a/flight/net/Request.php
+++ b/flight/net/Request.php
@@ -164,9 +164,7 @@ class Request {
         }
 
         // Get the requested URL without the base directory
-        if ($this->base != '/' && strlen($this->base) > 0 && strpos($this->url, $this->base) === 0) {
-            $this->url = substr($this->url, strlen($this->base));
-        }
+        $this->url = self::getVar('QUERY_STRING');
 
         // Default url
         if (empty($this->url)) {


mikecao:
The problem is that there are hundreds of ways to set up the rewrite rules on any server. Your change completely gets rid of the base directory check, which is the common use case. It may work for you but definitely not for everyone else.

Sigill:
Actually, this might not be the proper way to do it, but being able to run Flight from a subdirectory would be a great plus for security.

pine3ree:
@Sigill 
Hello, i believe the proper thing to do in your case is to make "public/" the document root.
vendor/, config/ and other sensible or library directory/file should always dwell outside the web server docroot. You should keep only the front controller script and the public assets inside the docroot.
You just need to include the autoloader path relative to the front script.

if you have a dir structure like the following

path/to/DOCROOT/subdir/index.php
path/to/DOCROOT/subdir/.htaccess
path/to/DOCROOT/subdir/css
path/to/DOCROOT/subdir/js
path/to/DOCROOT/subdir/img/
...

because for some reason you need your application in a subdir/ (and a suburl from the browser point of view) the vendor directory should be here:

path/to/vendor

and NOT here

path/to/DOCROOT/vendor

NOR here

path/to/DOCROOT/subdir/vendor

kind regards,
maks feltrin
pine3ree

Sigill:
I agree with you, but you might not want/be able to make public/ the document root of the webserver.

One of the things I like about PHP is that with Apache, most of the time, you can have your website running by just copying it somewhere in your documentroot (and, since Flight relies on mod_rewrite, defining the RewriteBase in the .htaccess). This is not the ideal case, but for testing purpose, it is quite convenient.

How will you write and configure a Flight app to work with Apache in both the case discribed above and in the case where you can make public/ the documentroot of your virtualhost?

pine3ree:
@Sigill 
In that case you need to edit .htaccess where the front script index.php resides.

```
RewriteEngine On
RewriteBase '/subdir/'
RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{REQUEST_FILENAME} !-d
RewriteRule ^(.*)$ index.php [QSA,L]
```

index.php

``` php
<?php
require __DIR__ . '/../../vendor/autoload.php';

use flight\Engine;

$app = new Engine();

$app->route('/', function() {
    echo 'home!';
});

$app->route('/hello-1', function() {
    echo 'hello 1!';
});

$app->route('/hello-2', function() {
    echo 'hello 2!';
});

$app->start();
```

i tested this configuration and the following routes are handled correctly:
/subdir
/subdir/hello-1
/subdir/hello-2

As you noticed from the 1st line (require **DIR** . '/../../vendor/autoload.php';)
vendor/ directory is still kept outside docroot. Do no rely too much on apache rewrite to block access to paths inside the docroot, even the slgihtest misconfiguration or incompatibility with base configuration from the hosting company can break .htaccess security settings.

kind regards

Sigill:
The problem is a bit more complex: I actually want the `public/` directory to be transparent to the user.

While trying to build an example to clarify the problem, I think I've found the solution (highly inspired by the one used in [CakePHP](http://book.cakephp.org/3.0/en/installation.html#apache)).

Here it is: https://github.com/Sigill/flight-public-docroot.

When Apache is configured to directly serve the `public/` directory, the `public/app.php` script will handle the request as usual (unless the request is for a static file, of course).

When the root directory of the website is served, the main `.htaccess` will internally redirect the request to the `public/` directory. Static files will be served from here as usual, but other requests cannot be handled by the `public/app.php` script (otherwise, the extra `public/` in the path will prevent Flight to match the routes properly). The trick is to have them handled by the `app.php` script at root level (which only includes the `public/app.php` script). The only thing to do is to set the `RewriteBase` directive in both `.htaccess` files.

PR #212, opened 2015-10-22
bages + travis-ci + scrutinizer

tamtamchik:
**To make it work correctly you will need to login and enable service for this repo on https://travis-ci.org and https://scrutinizer-ci.com (with github account)**

Added some tranding wrappers for packages like Travis-CI autotests and Scrutinizer code metrics.
Benefits:
- badges immidiately show all related to package metrics, so users will see what they are about to use, how many other users use this product and can they trust the code... 
  
- while commits and PR it's easy to see status, and check that everything is OK, and nothing would be broken.
  

Also small tweeks to running tests via composer.
Just close PR if you don't worry about all this.

Thanks.


ReenExe:
:+1:
Why so long open?

fadrian06:
Hi there! Instead of adding Travis CI, we'll be implementing GitHub Actions for this pull request.

PR #221, opened 2015-12-01
set only references to variables in views

edam:
set only references to variables in the view, rather than taking a copy of objects (which may be non-trivial)

### DIFF
--- a/flight/template/View.php
+++ b/flight/template/View.php
@@ -60,14 +60,14 @@ class View {
      * @param mixed $key Key
      * @param string $value Value
      */
-    public function set($key, $value = null) {
+    public function set($key, &$value = null) {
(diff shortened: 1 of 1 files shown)


mikecao:
PHP uses a copy-on-write optimization so passing by reference is actually slower than passing values. Plus template variables should be separate values anyways.

edam:
Are you sure? I can think of use cases for views to have their own copy of an object and also a reference (e.g., to a non-trivial object). Perhaps both methods would be useful. Would you accept a different patch that adds a separate addRef() method? Or would there be some way of reusing the function for both purposes? Or are you not interested in this functionality at all?

mikecao:
As of PHP 5, all objects are passed and assigned by reference. You don't need to indicate it in the method.

PR #246, opened 2016-03-10
Updated README.md

tamtamchik:
Added example for routing object methods to readme, because a lot of issues created because of this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,9 @@ class Greeting
     }
 }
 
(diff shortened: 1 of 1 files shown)


mikecao:
I think it's bad practice for users to declare objects in the routes. A better example would be if it was declared already, `array($greeting, 'hello')`.

tamtamchik:
@mikecao updated.

PR #252, opened 2016-03-24
Note regarding usage of flight in a subdirectory

hitautodestruct:
Thanks to this question and answer http://stackoverflow.com/questions/20890210/flight-php-routing-from-subdirectory


mikecao:
Can you change /flight/ to /subdir/? I don't want to give the impression that the name is flight specific.

hitautodestruct:
Done! :-)

PR #254, opened 2016-03-28
Develop

neko2891:
Please take it as your reference.Please refer to it.

getParam() is reference from
http://stackoverflow.com/questions/2081894/handling-put-delete-arguments-in-php

### DIFF
--- a/flight/Flight.php
+++ b/flight/Flight.php
@@ -50,6 +50,9 @@
  * @method  static void etag($id, $type = 'strong') Performs ETag HTTP caching.
  * @method  static void lastModified($time) Performs last modified HTTP caching.
  */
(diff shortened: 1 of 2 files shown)


mikecao:
1) Flight is supposed to be in the global namespace. The Engine class is already under a namespace.

2) PUT/DELETE should not be automatically parsed. PUT data can be anything, including a large file. Since `php://input` can only be read once, you now can't access the data anymore. Users should be using POST if they need to pass parameters. DELETE should also not pass parameters. Since it should rely on the url you are sending DELETE to, like `/user/123`.

neko2891:
OK, I got it. thanks for your response.

PR #278, opened 2016-07-18
Route support '(' and ')', such as: @name:(foo|bar)

lin-credible:


mikecao:
Adding support for parentheses would break the regular expression matching because you are introducing an unknown number of match groups.

PR #279, opened 2016-07-22
Second argument to get() function

catsAND:
Add second argument to get() function. Its need when we registering with set() function two-dimensional array.


mikecao:
The get method is supposed to get data by key. It's not supposed to deal with nested data. Also PHP arrays are n-dimensional which require an unknown number of parameters.

Issue #290, opened 2016-09-17
Can json_encode options to be set?


wwtg99:
In the `Flight::json()` function, options of json_encode can not be set such as `JSON_UNESCAPED_UNICODE`,

```
public function _json($data, $code = 200, $encode = true, $charset = 'utf-8') {
        $json = ($encode) ? json_encode($data, JSON_UNESCAPED_UNICODE) : $data;
        $this->response()
            ->status($code)
            ->header('Content-Type', 'application/json; charset='.$charset)
            ->write($json)
            ->send();
}
```

Can this option be set manually by passing parameters or load from config?
Thank you!

ghost:
@wwtg99 This enhancement has been merged, are you happy with the change?

PR #294, opened 2016-10-11
#290 - JSON methods now support encode option/depth

ghost:
JSON methods now support encode option/depth, added missing arguments to json/jsonp method documentation, _json/_jsonp methods now list arguments on individual lines to remain PSR-2 compliant

This adds two new arguments instead of refactoring. In future, should flight decide to release a 2.x.x version number, we should _perhaps_ look at breaking backwards compatibility and replacing the `$encode` argument with an array of encode options whereby the following is true:

`$encode == false` ? Do not encode
`$encode === array() || $encode == true` ? Encode
`isset($encode['option'])` ? Pass option to json_encode parameter 2
`isset($encode['depth'])` ? Pass depth to json_encode parameter 3

Enhancement requested in: https://github.com/mikecao/flight/issues/290

### DIFF
--- a/README.md
+++ b/README.md
@@ -887,8 +887,8 @@ Flight::error($exception) // Sends an HTTP 500 response.
 Flight::notFound() // Sends an HTTP 404 response.
 Flight::etag($id, [$type]) // Performs ETag HTTP caching.
 Flight::lastModified($time) // Performs last modified HTTP caching.
-Flight::json($data, [$code], [$encode]) // Sends a JSON response.
-Flight::jsonp($data, [$param], [$code], [$encode]) // Sends a JSONP response.
+Flight::json($data, [$code], [$encode], [$charset], [$encodeOption], [$encodeDepth]) // Sends a JSON response.
+Flight::jsonp($data, [$param], [$code], [$encode], [$charset], [$encodeOption], [$encodeDepth]) // Sends a JSONP response.
 ```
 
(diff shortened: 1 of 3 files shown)


mikecao:
Thanks, great work!

josevh:
Hi all, this breaks PHP 5.3 compatibility.

I get the following 500 error in response tested on PHP 5.3.3.:
`json_encode() expects at most 2 parameters, 3 given (2)`

Depth was added in 5.5 per: [php.net](http://php.net/manual/en/function.json-encode.php)

I am not sure if the aim is to stay 5.3 compatible.
If not, the README should be updated.

Did I miss something? Should I open an issue?

mikecao:
@josevh Thanks for pointing it out. Yes Flight should be 5.3 compatible. I'll have to roll back the change.

ghost:
@mikecao @josevh By the look of it, we can simply remove the depth addition and leave the "options" values in there. Options was added in 5.3.0. The deeper "options" were added in 5.3.3/5.4.0, but by default we can support all "options" by just passing in an int, as we are now.

I think that's the best compromise, as it maintains full 5.3.0 compatibility and also fixes the original issue.

mikecao:
Fixed in 9c8d492799da73690ae0c370d8cd685148ef09f5.

tamtamchik:
Wow! New minor version! 🎉

PR #295, opened 2016-10-12
Attempting to execute unmapped methods now triggers Exceptions

ghost:
Previously, attempting to execute a method which had not been registered in the Engine's dispatcher would return a null because without a valid callable method, `Dispatcher::invokeMethod()` eventually falls through to `Loader::load()` which returns null by default if the method isn't defined in the classes list.

This removes that behaviour by throwing exceptions when undefined methods are accessed in the  Engine's dispatcher which are not otherwise mapped.

It also has the added benefit of no longer incorrectly instantiating classes when there is a crossover between a defined Flight internal class name and an incorrectly accessed method name - this undefined behaviour could be very confusing to developers.

Issue enhancement reference: https://github.com/mikecao/flight/issues/203

### DIFF
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -272,6 +272,16 @@ class Engine {
         $this->loader->addDirectory($dir);
     }
 
+    /**
+     * Checks for mapped event within the engine dispatcher.
+     *
+     * @param $name
+     * @return bool
+     */
+    public function isEvent($name) {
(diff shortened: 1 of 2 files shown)


mikecao:
Some minor changes in b30d5e5fa2de08c3e91c1165ee226fd83f29ba9b. I had to move the logic to the Engine level because some people may use the app instance instead of the global static class. Also it was failing some tests.

PR #301, opened 2016-11-20
Fix redirect method due to a mistake removing backslashes

yassineaddi:
The problem is that if 'flight.base_url' setting is set to 'http://localhost/flight', then:
$url = 'http:/localhost/flight';
Notice that one backslash has been removed.


mikecao:
Thanks!

yassineaddi:
You're welcome!

PR #304, opened 2016-11-27
Request data utf-8 encoding

mark-szabo:
UTF-8 encoding the data object in constructor

### DIFF
--- a/flight/util/Collection.php
+++ b/flight/util/Collection.php
@@ -26,7 +26,7 @@ class Collection implements \ArrayAccess, \Iterator, \Countable {
      * @param array $data Initial data
      */
     public function __construct(array $data = array()) {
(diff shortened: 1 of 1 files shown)


ghost:
@szmaorka I'm not sure that this should be merged, this feels implementation specific and not something we should impose on every user of Flight.

It's likely that we have users who's applications currently expect to be receiving data in UTF-8, UTF-16, ASCII, UTF-EBCDIC, etc and this could be a breaking change for their API. Encoding to UTF-8 is trivial if your application needs it and it's my belief that this feature should remain in the user space code, though of course @mikecao has the final word.

mikecao:
Agreed, we should not be enforcing any kind of encoding on the user.

PR #329, opened 2017-07-17
Calculate proper request size when using mbstring overloading

Jeanvanberlo:
strlen doesn't calculate proper size when using multibyte strings.


mikecao:
mbstring functions are not included in the default PHP installation.

Jeanvanberlo:
You are of course correct, I've added a check to see if mbstring is loaded, otherwise use the core string functions. This way it should work properly with and without mbstring

mikecao:
The original strlen already returns total number of bytes, which is what the content-length header is supposed to represent. I believe your solution returns the number of characters, which is not correct.

Jeanvanberlo:
Strlen assumes 1 char is 1 byte. This is not correct for multibyte strings. mb_strlen($utf8_string, 'latin1'); works because latin1 only contains byte chars, this will then actually output the proper byte size for multibyte strings.

mikecao:
Both `strlen('€')` and `mb_strlen('€','latin1')` return 3 bytes.

Jeanvanberlo:
Because you probably haven't enabled the mbstring overload.
set this in the php.ini:
mbstring.func_overload = 2
and you results will be 1 byte for `strlen('€')` and 3 bytes for `mb_strlen('€','latin1')`

Jeanvanberlo:
While reading up on this I noticed this does get deprecated in php 7.2. Nonetheless not everyone will move to 7.2 in a short timeframe. More information can be found here (also on strlen when overloaded): http://markmail.org/message/vckycqqppwxhp2gk

mikecao:
Right, I don't have mbstring overloading enabled, which would be the default. So by default, `strlen` works correctly. Enabling it would have `strlen` call `mb_strlen` to get the wrong value, and again you are back at the issue of having mbstring installed in the first place.

It sounds like PHP really messed up in allowing the overload in the first place, since you can't guarantee what strlen will return on different environments.

Your solution is on the right track. If mbstring is installed I think we should use it to avoid the overloading issue. But your pull request has too many code styles changes for me to accept, so I will provide the fix separately.

Jeanvanberlo:
Yeah it's not the best of php. Probably for the best it got deprecated too. I understand you won't accept the pull, sorry my auto psr formatting changed the style, and I committed before noticing. Thanks for implementing this fix!

mikecao:
Fixed in 1c8d06cc42dcb04355984d561ce9fbbcf0bb9ef7.

PR #347, opened 2017-12-05
[BREAKING] Switch Flight to PSR-4 standards.

jakewtaylor:
This PR converts flight to use PSR-4 standards. In order for this to work, I've **dropped support for using Flight _without_ Composer**. 

Changes:
- Renamed directories to match class names:
  - `flight/core` -> `flight/Core`
  - `flight/net` -> `flight/Net`
  - `flight/template` -> `flight/Template`
  - `flight/util` -> `flight/Util`
- Changed `composer.json`s `autoload` property to use PSR-4 instead of 2 files:
  ```JSON
  "autoload": {
      "psr-4": {
          "Flight\\": "flight/"
      }
  }
  ```
- Changed all `use` statements and `namespace` statements to be PascalCase
- Changed PHPDoc blocks to use the right namespaces.

I've changed the tests, and they pass, but wouldn't mind a second pair of eyes to check them as I'm not good with tests. 😄

Obviously this is a breaking change, as it drops support for using the framework without composer.

I needed this as Flight was not playing ball with my application that uses PSR-4, and I was getting many errors because the autoloading wasn't working well.

Leaving all the autoloading to composer means there's one less thing to worry about here. I assume we can either remove or change some of the `Flight\Core\Loader` class as we shouldn't need to do any autoloading now, but I don't fully understand the inner workings of the framework, or PHP autoloading, so will have to leave that for now.

Let me know your opinions! 😋


jakewtaylor:
Not sure why the build is failing... let me know how to fix and I will 😃

jakewtaylor:
@mikecao any thoughts on this?

mikecao:
@jakewtaylor This is too much of a breaking change for me to pull. Not everyone uses or is familiar with Composer and I'd rather keep that option open.

I'm not sure what problems you were running into, but the Flight autoloader does allow you to add directories to look for files if you ever need to do that.

jakewtaylor:
@mikecao, that's fair enough.

I was running into problems because I wasn't using Flights autoloading, but using Composers autoloading because I was using lots of other composer packages in my project.

The error being thrown was because the Flight class was being declared more than once, which I assume was because I was doing `use Flight\Flight;` in multiple classes to access methods. I don't know a huge amount about autoloading, but I assume Flights autoloader was importing the main class every time the `use` statement was used, or colliding with Composers autoload.

Is there any way to disable Flights built-in autoloading and only use Composers?

If you like the PSR-4 format, and the only reason for not merging is backwards compatibility, maybe it could be released as v2 at some point, making it obvious that this is a breaking change.

mikecao:
You can have multiple autoloaders running at once. They just run in the order they are declared and if none of them can find the file you're declaring you will get an error. They really shouldn't conflict since they look for files using independent methods. But anyways, you can disable Flight's autoloader by passing `false` to the autoload method. Take a look at `autoload.php`.

Yes at some point, in a v2 release I will consider doing more breaking changes and adopting more PSRs.

bubach:
Nice, was leaning towards forking it myself and bringing it up to speed with 2019. 👍

Conver:
I like the idea about a v2 for sure, I use FlightPHP for ALL my php apps.

I can understand many people like using Composer, I can see many benefits using it,
- But I hate using Composer in small to semi small projects, so removing the manually option would be a very downgrade for me and I would have to find another framework to use with those smaller projects. So I deeply hopes that manually options will stay

evansharp:
Is this v2 thing a moonshot at this point?

jakewtaylor:
Hey folks, doing some personal house cleaning - closing this as it's been stale for a while. Feel free to reopen if you like.

PR #351, opened 2018-01-02
Stop method should exit

sabas:


levinside:
You can do it like this:
`Flight::after('stop', function(){ exit; });`

=)

mikecao:
Stop doesn't exit on purpose so that you can do what @levinside mentioned, which is run post filters.

PR #355, opened 2018-02-02
Removed extraneous ->write($url) call on redirect that was causing Sa…

jqs:
…fari to bugger out qwhen using HTTP2, which it tends to handle more strictly.

Notes:
The line above writes the URL to redirect to for the 303 then that line writes it again.
The result is a multi-line http header which causes safari to freak out when in http2 mode.
https://blog.christopherburg.com/2017/10/18/safari-11-multiline-http-headers-and-nsposixerrordomain100/
blog.christopherburg.com
Safari 11, Multiline HTTP Headers, and NSPOSIXErrorDomain:100. at A Geek With Guns
Chronicling the depravities of the State

### DIFF
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -437,7 +437,6 @@ class Engine {
             ->clear()
             ->status($code)
             ->header('Location', $url)
-            ->write($url)
             ->send();
     }
 


jqs:
Hopefully this can be merged in so our production environment can be back to using base packages.

Cheers.

PR #358, opened 2018-02-24
Update README.md

paxperscientiam:
Not sure if it's just a composer thing, but class registration does not work for me unless I use the FQCN.


mikecao:
I don't think it's a composer only thing. You have to provide the full name because that's how the autoloader looks for things.

PR #380, opened 2018-10-26
Initial

frantisekpetko:
Hi, I maked a pull request (namespace at Flight class), thx for solving my issue. :)


mikecao:
The Flight class can't be under a namespace. It needs to be global for ease of use.

PR #382, opened 2019-01-03
add ability to call non-static methods

nick-mad:
Now you can call non-static methods. When there are many controllers on the web site, you need to initiate everything, and now one of them has been initialized.
You can do it:
`class Greeting
{
    public function __construct() {
        $this->name = 'John Doe';
    }

    public function hello() {
        echo "Hello, {$this->name}!";
    }
}

Flight::route('/', array('Greeting', 'hello')); `

### DIFF
--- a/flight/core/Dispatcher.php
+++ b/flight/core/Dispatcher.php
@@ -191,7 +191,15 @@ class Dispatcher {
         list($class, $method) = $func;
 
         $instance = is_object($class);
-		
+
+        if (!$instance && method_exists($class, $method)) {
+            $methodChecker = new \ReflectionMethod($class, $method);
+            if (!$methodChecker->isStatic()) {
+                $class = new $class;
+                $instance = is_object($class);
+            }
+        }
+
         switch (count($params)) {
             case 0:
                 return ($instance) ?


mikecao:
It looks like this would create a new instance of the class every time it runs and you have no way of controlling that instance. The proper way to do it would be to instantiate the instance yourself and then pass it in to the route.

nick-mad:
If you create an instance of a class yourself, then with a large number of routes you need to initiate all classes that are involved in all routes. If the constructors of these classes have dependencies on other classes, this automatically initiates them. As a result, when processing a single route, we initiate all the classes that are hopped up in the routes. When using the correction proposed by me, only the class that got under the rules of the route will be initiated.

Take for example a small application.


```
class Home
{
    public function __construct()
    {
        var_dump('Instantiate Home Controller and dependency');
    }

    public function index()
    {
        echo 'Home page';
    }
}

class Categories
{
    public function __construct()
    {
        var_dump('Instantiate Categories Controller and dependency');
    }

    public function showList()
    {
        echo 'Categories list';
    }
}

class Products
{
    public function __construct()
    {
        var_dump('Instantiate Products Controller  and dependency');
    }

    public function showList()
    {
        echo 'Products list';
    }
}

class Orders
{
    public function __construct()
    {
        var_dump('Instantiate Orders Controller and dependency');
    }

    public function showList()
    {
        echo 'Orders list';
    }
}
```
if instantiate the instance yourself and then pass it in to the route.
```
require 'vendor/mikecao/flight/flight/Flight.php';

$home = new Home();
$categories = new Categories();
$products = new Products();
$orders = new Orders();

Flight::route('/', array($home, 'index'));
Flight::route('/categories', array($categories, 'showList'));
Flight::route('/products', array($products, 'showList'));
Flight::route('/orders', array($orders, 'showList'));

Flight::start();
```

http://localhost:8000/
We see  that all classes involved in routes are initiated.

```
flight-demo.php:8:string 'Instantiate Home Controller and dependency' (length=42)
flight-demo.php:21:string 'Instantiate Categories Controller and dependency' (length=48)
flight-demo.php:34:string 'Instantiate Products Controller  and dependency' (length=47)
flight-demo.php:47:string 'Instantiate Orders Controller and dependency' (length=44)

Home page
```

If the object is initiated by itself

```
require 'vendor/mikecao/flight/flight/Flight.php';

Flight::route('/', array('Home', 'index'));
Flight::route('/categories', array('Categories', 'showList'));
Flight::route('/products', array('Products', 'showList'));
Flight::route('/orders', array('Orders', 'showList'));

Flight::start();
```
http://localhost:8000/
We see that only one class is initiated in the same route.

```
flight-demo.php:7:string 'Instantiate Home Controller and dependency' (length=42)

Home page
```

As a result, if there are a large number of routes, it is necessary to initiate all classes of controllers and all their dependencies.

The main thing is that backward compatibility is not violated, everyone who has previously passed an object to the route or called static methods can continue to use them. It simply complements or expands the framework.

mikecao:
What you're doing is introducing a side-effect, which is bad practice. The route expects an actual callable so  there shouldn't be any magic happening.

But you can still do what want by creating your own function and passing that to the routes.

```php
function controller($class, $method) {
  $obj = new $class();
  $obj->$method();
}
```

Then define your routes

```php
Flight::route('/', function() { controller('Home', 'index'); });
Flight::route('/categories', function() { controller('Categories', 'showList'); });
// etc
```

That's the simplest way. But if you want to get fancy you can create an array of routes to classes and have the controller function look it up.

PR #384, opened 2019-01-10
Allow easy routing of URLs containing Cyrillic letters.

spytheman:
I want the ability to route URLs containing Cyrillic letters. 
For example: /категория/цветя  (in english: /category/flowers ). 

As it is, the url gets encoded in the request->url as '%2F%D0%BA%D0%B0%D1%82%D0%B5%D0%B3%D0%BE%D1%80%D0%B8%D1%8F%2F%D1%86%D0%B2%D0%B5%D1%82%D1%8F'

The patch in this pull request, gives flight the ability to match Cyrillic urls.

### DIFF
--- a/flight/net/Router.php
+++ b/flight/net/Router.php
@@ -78,8 +78,9 @@ class Router {
      * @return Route|bool Matching route or false if no match
      */
     public function route(Request $request) {
+        $url_decoded = urldecode( $request->url );
         while ($route = $this->current()) {
-            if ($route !== false && $route->matchMethod($request->method) && $route->matchUrl($request->url, $this->case_sensitive)) {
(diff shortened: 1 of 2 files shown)


raunsbaekdk:
Hi @mikecao 

Apparently this breaks routing with forward slash in string.

Consider the following code:

```
Flight::route('GET /welcome/@string', function($string) {
  print_r($string);
})
```

Calling the webserver with cURL:
curl "http://localhost/test+q%2Fa" now gives 404.

Line 82 and 83 changed the behaviour of how the URL matches.

The URL in the match before would be:
/welcome/test q%2Fa

Now it is:
/welcome/test q/a

And thus fails in the matching process.

Do you have a workaround for this problem?

PR #385, opened 2019-01-19
 Update execute() method

lorenzosanzari:
In the official definition of Flight's routes, it is allowed to express the class callables as an array consisting of

[string, string] // static class 

or

[object , string] // object instance

In the second case, already in the definition we have to write:

[new MyClass (), 'myMethod']

which immediately causes the class constructor execution, and this can cause problems.

With the code proposed by me, however, the object instance will be created only when the route will actually be invoked.
So, even for non-static classes, we can simply write:

['myClass', 'myMethod']

avoiding the immediate execution of the constructor.

### DIFF
--- a/flight/core/Dispatcher.php
+++ b/flight/core/Dispatcher.php
@@ -139,14 +139,25 @@ class Dispatcher {
      * @throws \Exception
      */
     public static function execute($callback, array &$params = array()) {
-        if (is_callable($callback)) {
-            return is_array($callback) ?
-                self::invokeMethod($callback, $params) :
-                self::callFunction($callback, $params);
-        }
-        else {
-            throw new \Exception('Invalid callback specified.');
-        }
+
+	if (is_array($callback) && is_string($callback[0]) && isset($callback[1])) {
+	    $classname = $callback[0];
+	    $method = $callback[1];
+	    if (class_exists($classname)) {
+		$r_method = new ReflectionMethod("$classname::$method");
+		if (!$r_method->isStatic())  //is not a static method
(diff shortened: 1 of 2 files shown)


mikecao:
You are proposing the same feature as this PR, https://github.com/mikecao/flight/pull/382. It has the same problem because it's not a real callable. You can solve the execution issue by passing in a wrapper function as a callback. See the comments in the PR.

lorenzosanzari:
I do not understand why it should be a problem that it is not a "real callable".
It is a callable "DEFINITION", (as it happens in most frameworks), and sincerely I don't see the problem of "control" the instance. (Also with your controller() function there is not control on instance: it makes the same thing).
I am using this extended method in a bolierplate based on Flight  (created by me) and it works very well, without the undesired effect of the builder's execution.
But I leave you the choice to integrate or not the pull request. :-)

lorenzosanzari:
I you want, we can also do something like this:

....

if (is_array($callback) && is_string($callback[0]) && isset($callback[1])) {
	    $classname = $callback[0];
	    $method = $callback[1];
	    if (class_exists($classname)) {
		$r_method = new ReflectionMethod("$classname::$method");
		if (!$r_method->isStatic())  //is not a static method
		    $callback[0] = new $callback[0](); //instantiate object on the fly (Lorenzo Sanzari)
      >>>>    Flight::setController($callback[0]); //to have instance in Flight context
	    } else {
		throw new \Exception('The class ' . $callback[0] . ' does not exists!');
	    }
	}

....

n0nag0n:
If this is still an issue you'd like to address, go ahead an open up an issue so it can be discussed further.

Issue #396, opened 2019-07-23
Flight::request()->scheme return SERVER_PROTOCOL and not REQUEST_SCHEME


hlidotbe:
`Flight::request()->scheme` does not return _http_ or _https_ as documented but the server protocol (which is almost always HTTP/1.1 and not the scheme).

`REQUEST_SCHEME` should be used instead of `SERVER_PROTOCOL`. `HTTP_X_FORWARDED_PROTO` should be handled if present instead of `REQUEST_SCHEME`

pwFoo:
The PR is merged. So issue is fixed? Could be closed?

hlidotbe:
Yes indeed

PR #397, opened 2019-07-23
Fix request scheme

hlidotbe:
As mentionned in #396, the request scheme was not as documented,
returning `SERVER_PROTOCOL` instead.

The `getScheme` function is now used to handle common cases (HTTPS,
FORWATED_PROTO, ...).

`request->secure` is also based on the scheme now as `$_SERVER['HTTPS']`
is unreliable for this purpose.


mikecao (approved):
Thanks!

PR #406, opened 2019-12-28
Implement Cross Origin Resource Sharing

aleferri:
- added optional parameter cors_allowed to Flight::route
- propagated this optional parameter to net\Route constructor
- check before the dispatch of the selected route if the route allow cors, if so send the required headers before processing it


aleferri:
Will redo it better in a moment

PR #419, opened 2020-04-29
Middleware backward compatible implementation

aleferri:
- Removed $pass property on net\Route
- Removed $pass parameter on net\Router::map
- Added $config property on net\Route
- Added $config parameter on net\Router::map
- Made dispatchRoute replaceable by the user
- Added optional Middleware Routing Handler (LayersStack and LayersIterator), in the end the router call the original _dispatchRoute
- In Flight::start check $config['pass_route'] instead of $pass (now it doesn't exists anymore)
- All tests updated to PHPUnit 7
- Bumped required php version from 5.2 to 5.6
- Bumped required phpunit version from 4.8 to 7
- Updated the readme.md to show how to use the new Middleware enabled implementation of routing


Conver:
I don't really see a reason for this PR, can you clarify why this PR should be merged?

n0nag0n:
If this is still an issue you'd like to address, go ahead an open up an issue so it can be discussed further.

flight/core/Dispatcher.php
```diff
@@ -46,7 +46,7 @@ public function run($name, array $params = array()) {
         }
 
         // Run requested method
-        $output = $this->execute($this->get($name), $params);
+        $output = $this->get($name)(...$params);
```
n0nag0n, line 49: I actually like a lot of the cleanup you did in this file. I might borrow some of the ideas for the middleware feature I'm building into Flight

PR #456, opened 2021-12-17
PHP 8.1: Type declare length as int

jamalo:
The following error is thrown when length is set to '10' instead of 10:
`TypeError: Cannot assign string to property flight\net\Request::$length of type int`

### DIFF
--- a/flight/net/Request.php
+++ b/flight/net/Request.php
@@ -147,7 +147,7 @@ final class Request
                 'scheme' => self::getScheme(),
                 'user_agent' => self::getVar('HTTP_USER_AGENT'),
                 'type' => self::getVar('CONTENT_TYPE'),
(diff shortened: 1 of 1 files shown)


flight/net/Request.php
```diff
@@ -147,7 +147,7 @@ public function __construct(array $config = [])
                 'scheme' => self::getScheme(),
                 'user_agent' => self::getVar('HTTP_USER_AGENT'),
                 'type' => self::getVar('CONTENT_TYPE'),
-                'length' => self::getVar('CONTENT_LENGTH', 0),
+                'length' => (int)self::getVar('CONTENT_LENGTH', 0),
```
mu1f407:
```suggestion
                'length' => (int) self::getVar('CONTENT_LENGTH', 0),
```
Better with space after cast.

PR #515, opened 2024-01-12
Fixed tests and lint problems in windows

fadrian06:


flight/template/View.php
```diff
@@ -130,6 +130,7 @@ public function render($file, $data = null)
         $this->template = $this->getTemplate($file);
 
         if (!file_exists($this->template)) {
+            $this->template = self::normalizePath($this->template);
```
n0nag0n: So this will fix the path....but then it throws an exception so nothing is really fixed.

fadrian06: Do you mean about empty strings?

fadrian06: Do you mean about empty strings?

tests/ViewTest.php
```diff
@@ -117,4 +120,27 @@ public function testENoNeedToEscape() {
...
+    public function testNormalizePath(): void
+    {
+        $viewMock = new class extends View {
+            public static function normalizePath(string $path, string $separator = DIRECTORY_SEPARATOR): string
+            {
+                return parent::normalizePath($path, $separator);
+            }
+        };
+
+        self::assertSame(
```
n0nag0n: Not a huge deal but the rest of the file uses `$this->` instead of `self::` One thing I've learned that is really important is to keep the codebase as consistent as possible so it doesn't look like "Oh that's definitely a spot fadrian06 coded in". When I code in this framework, even though it drives me nuts, I keep all the class vars lowercase. I would much rather do `$Router = Flight::router();` but in order to keep it consistent, I do `$router = Flight::router();` simply because it's keeping it consistent.

fadrian06: oh sorry, I hadn't noticed

n0nag0n (changes requested):
So if you remove the static typing and then correct L111 to be $normalized_template_path_for_error then I think this is good to go.

n0nag0n (approved):

PR #516, opened 2024-01-13
fixed alias issue, levenshtein recommendations and coverage-check

n0nag0n:


krmu (approved):
Works like it supposted to work.
image

PR #527, opened 2024-01-23
Removed all typehints to PHP <7.4 compatibility

fadrian06:
image
image
image

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2011 Mike Cao <mike@mikecao.com>
+Copyright (c) 2024 Mike Cao <mike@mikecao.com>
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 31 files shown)


n0nag0n (commented):
So many changes but seems like it'll get the job done. We do have a lot of unit tests in place that are all passing. What do you think @krmu ?

krmu:
> So many changes but seems like it'll get the job done. We do have a lot of unit tests in place that are all passing. What do you think @krmu ?

I will test it out.

fadrian06:
For a moment I was surprised when I changed the © haha... I know it's a massive change but it resolves inconsistencies in which some classes could work in <7.4 and others don't, so removing all the typehint not only maintains consistency but we also support php <7.4... We would have to test until which version works correctly, but I think in 7.0 it would work

krmu (approved):
Testing worked, but I would like to before merging move to discussion at matrix.

n0nag0n:
We are going to pass on this pull request and instead move forward with static typing on everything as it is stated that this project is 7.4 or greater. We are not going to support older versions of PHP.

PR #531, opened 2024-01-25
PHPCS code formatting problems solved

fadrian06:


n0nag0n (approved):
lgtm, but I don't know why it beautified so differently for you than it did me unless you are forcing sublime to enforce coding standards outside of PSR12

fadrian06:
I have the EditorConfig plugin in Sublime That's why the LSP always respects the 4 spaces and replaces all the tabs that are left there.

fadrian06:
in the source code it does not automatically format unless you commit only the formatting...in this PR I did all the formatting by hand to reduce the size of the lines  since phpcs complains about lines longer than 120 characters

fadrian06:
finally I applied phpcbf but it didn't do anything because almost everything was already formatted

PR #534, opened 2024-01-28
fixed bug with root alias

n0nag0n:


n0nag0n:
This would return `''` for an alias if it was the `/` path

fadrian06 (approved):
Does your editor use tabs instead of spaces? I have already noticed it from several commits

n0nag0n:
Gah, yeah I keep switching it. Lemme fix it.

fadrian06:
It's strange, that's why I set the .editorconfig to follow the spaces by tabs, I also preferred tabs

fadrian06:
It's strange, that's why I set the .editorconfig to follow the spaces by tabs, I also preferred tabs, but on github it makes it difficult to read the diff because each tab gives 4 spaces and I have 2

fadrian06:
It's strange, that's why I set the .editorconfig to follow the spaces by tabs, I also preferred tabs, but on github it makes it difficult to read the diff because each tab gives 4 spaces and I have 2

PR #535, opened 2024-01-30
fix testStaticFunctionMapping to actually map a static method

lubiana:
The test was violating phps strict_types, did not test what the name suggested, and was failing.


n0nag0n (approved):
Thanks so much!

PR #538, opened 2024-02-02
Dispatcher improvements.

fadrian06:
- `Dispatcher::set` method chaining, `$dispatcher->set(...)->set(...);`.
- `Dispatcher::hook` method chaining, `$dispatcher->hook(...)->hook(...);`.
- Added `Dispatcher::FILTER_BEFORE` and `Dispatcher::FILTER_AFTER` constants.
- Invalid filter types throws an E_USER_NOTICE.
- `$dispatcher->set('method1')->set('method1')` now throws an E_USER_NOTICE.
- Now `Dispatcher::execute` accept calls to instance methods statically.
```php
class MyController {
  function index( ) { }
  static function create( ) { }
}

Dispatcher::execute(['MyController', 'create']);
Dispatcher::execute(['MyController', 'index');
// Both callables now are valid if MyController doesn't receive constructor params.
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,3 +5,6 @@ indent_style = space
 indent_size = 4
 end_of_line = lf
 charset = utf-8
+
+[*.md]
+indent_size = 2
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -145,7 +145,7 @@ class Engine
         ];
 
         foreach ($methods as $name) {
-            $this->dispatcher->set($name, [$this, '_' . $name]);
+            $this->dispatcher->set($name, [$this, "_$name"]);
         }
 
         // Default configuration settings
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -32,7 +32,7 @@ use flight\net\Route;
  * # Routing
  * @method Route route(string $pattern, callable $callback, bool $pass_route = false, string $alias = '')
  * Routes a URL to a callback function with all applicable methods
- * @method void group(string $pattern, callable $callback, array $group_middlewares = [])
+ * @method void group(string $pattern, callable $callback, array<int, callable|object> $group_middlewares = [])
  * Groups a set of routes together under a common prefix.
  * @method Route post(string $pattern, callable $callback, bool $pass_route = false, string $alias = '')
  * Routes a POST URL to a callback function.
@@ -46,7 +46,7 @@ use flight\net\Route;
  * @method string getUrl(string $alias) Gets a url from an alias
  *
  * # Views
- * @method void render(string $file, array $data = null, string $key = null) Renders template
+ * @method void render(string $file, ?array $data = null, ?string $key = null) Renders template
  * @method View view() Gets current view
  *
  * # Request-Response
@@ -57,58 +57,41 @@ use flight\net\Route;
  * @method void redirect(string $url, int $code = 303)  Redirects the current request to another URL.
  * @method void json(mixed $data, int $code = 200, bool $encode = true, string $charset = 'utf-8', int $option = 0)
  * Sends a JSON response.
- * @method void jsonp(mixed $data, string $param = 'jsonp', int $code = 200, bool $encode = true, string $charset = 'utf-8', int $option = 0) Sends a JSONP response.
+ * @method void jsonp(mixed $data, string $param = 'jsonp', int $code = 200, bool $encode = true, string $charset = 'utf-8', int $option = 0)
+ * Sends a JSONP response.
  *
  * # HTTP caching
- * @method void etag($id, string $type = 'strong') Handles ETag HTTP caching.
+ * @method void etag(string $id, ('strong'|'weak') $type = 'strong') Handles ETag HTTP caching.
  * @method void lastModified(int $time) Handles last modified HTTP caching.
  */
 class Engine
 {
-    /**
-     * Stored variables.
-     * @var array<string, mixed>
-     */
-    protected array $vars;
+    /** @var array<string, mixed> Stored variables. */
+    protected array $vars = [];
 
-    /**
-     * Class loader.
-     */
+    /** Class loader. */
     protected Loader $loader;
 
-    /**
-     * Event dispatcher.
-     */
+    /** Event dispatcher. */
     protected Dispatcher $dispatcher;
 
-    /**
-     * If the framework has been initialized or not
-     *
-     * @var boolean
-     */
+    /** If the framework has been initialized or not. */
     protected bool $initialized = false;
 
-    /**
-     * Constructor.
-     */
     public function __construct()
     {
-        $this->vars = [];
-
         $this->loader = new Loader();
         $this->dispatcher = new Dispatcher();
-
         $this->init();
     }
 
     /**
      * Handles calls to class methods.
      *
-     * @param string $name   Method name
-     * @param array<int, mixed>  $params Method parameters
+     * @param string $name Method name
+     * @param array<int, mixed> $params Method parameters
      *
      * @throws Exception
-     *
      * @return mixed Callback results
      */
     public function __call(string $name, array $params)
@@ -120,7 +103,7 @@ class Engine
         }
 
         if (!$this->loader->get($name)) {
-            throw new Exception("{$name} must be a mapped method.");
+            throw new Exception("$name must be a mapped method.");
         }
 
         $shared = empty($params) || $params[0];
@@ -128,11 +111,11 @@ class Engine
         return $this->loader->load($name, $shared);
     }
 
-    // Core Methods
+    //////////////////
+    // Core Methods //
+    //////////////////
 
-    /**
-     * Initializes the framework.
-     */
+    /** Initializes the framework. */
     public function init(): void
     {
         $initialized = $this->initialized;
@@ -148,7 +131,8 @@ class Engine
         $this->loader->register('request', Request::class);
         $this->loader->register('response', Response::class);
         $this->loader->register('router', Router::class);
-        $this->loader->register('view', View::class, [], function ($view) use ($self) {
+
+        $this->loader->register('view', View::class, [], function (View $view) use ($self) {
             $view->path = $self->get('flight.views.path');
             $view->extension = $self->get('flight.views.extension');
         });
@@ -159,6 +143,7 @@ class Engine
             'render', 'redirect', 'etag', 'lastModified', 'json', 'jsonp',
             'post', 'put', 'patch', 'delete', 'group', 'getUrl',
         ];
+
         foreach ($methods as $name) {
             $this->dispatcher->set($name, [$this, '_' . $name]);
         }
@@ -192,15 +177,15 @@ class Engine
     /**
      * Custom error handler. Converts errors into exceptions.
      *
-     * @param int    $errno   Error number
-     * @param string $errstr  Error string
+     * @param int $errno Error number
+     * @param string $errstr Error string
      * @param string $errfile Error file name
-     * @param int    $errline Error file line number
+     * @param int $errline Error file line number
      *
+     * @return false
      * @throws ErrorException
-     * @return bool
      */
-    public function handleError(int $errno, string $errstr, string $errfile, int $errline)
+    public function handleError(int $errno, string $errstr, string $errfile, int $errline): bool
     {
         if ($errno & error_reporting()) {
             throw new ErrorException($errstr, $errno, 0, $errfile, $errline);
@@ -214,7 +199,7 @@ class Engine
      *
      * @param Throwable $e Thrown exception
      */
-    public function handleException($e): void
+    public function handleException(Throwable $e): void
     {
         if ($this->get('flight.log_errors')) {
             error_log($e->getMessage()); // @codeCoverageIgnore
@@ -226,7 +211,7 @@ class Engine
     /**
      * Maps a callback to a framework method.
      *
-     * @param string   $name     Method name
+     * @param string $name Method name
      * @param callable $callback Callback function
      *
      * @throws Exception If trying to map over a framework method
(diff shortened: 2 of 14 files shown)


fadrian06 (commented):
It works on my machine :D

krmu:
> It works on my machine :D

I will test it :)

.editorconfig
```diff
@@ -5,3 +5,6 @@ indent_style = space
 indent_size = 4
 end_of_line = lf
 charset = utf-8
+
+[*.md]
+indent_size = 2
```
n0nag0n, line 10: I know it's not really part of the commit, but just make sure that if the current indent size is 4, then this matches to be 4. If it is 2, then keep it 2.

flight/Engine.php
```diff
@@ -57,59 +57,41 @@
...
  */
-// phpcs:ignoreFile Generic.Files.LineLength.TooLong, PSR2.Methods.MethodDeclaration.Underscore
 class Engine
 {
-    /**
-     * Stored variables.
-     * @var array<string, mixed>
-     */
-    protected array $vars;
+    /** @var array<string, mixed> Stored variables. */
```
n0nag0n, line 69: phpcs would complain about this. Just be aware of that

flight/Engine.php
```diff
@@ -308,24 +300,27 @@ public function get(?string $key = null)
...
-     * @param mixed      $key   Key
-     * @param mixed|null $value Value
+     * @param string|iterable<string, mixed> $key
+     * Variable name as `string` or an iterable of `'varName' => $varValue`
+     * @param mixed $value Ignored if `$key` is an `iterable`
      */
     public function set($key, $value = null): void
     {
-        if (\is_array($key) || \is_object($key)) {
+        if (\is_iterable($key)) {
```
n0nag0n, line 309: Learn something new everyday :) Didn't know about that function!

flight/Engine.php
```diff
@@ -308,24 +300,27 @@ public function get(?string $key = null)
...
             foreach ($key as $k => $v) {
                 $this->vars[$k] = $v;
             }
-        } else {
-            $this->vars[$key] = $value;
+
+            return;
```
n0nag0n, line 314: I think this is ok. This seems to be a better way of handling it than it did before....I'll just need to keep it in mind for the unit tests.

flight/Engine.php
```diff
@@ -337,15 +332,16 @@ public function has(string $key): bool
...
-     * @param string|null $key Key
+     * @param ?string $key Variable name, if `$key` isn't provided, it clear all variables.
      */
     public function clear(?string $key = null): void
     {
         if (null === $key) {
             $this->vars = [];
-        } else {
-            unset($this->vars[$key]);
+            return;
```
n0nag0n, line 341: Same here

flight/Engine.php
```diff
@@ -554,55 +549,55 @@ public function _group(string $pattern, callable $callback, array $group_middlew
...
-     * @param bool     $pass_route Pass the matching route object to the callback
+     * @param string $pattern URL pattern to match
+     * @param callable $callback Callback function
+     * @param bool $pass_route Pass the matching route object to the callback
      */
-    public function _post(string $pattern, callable $callback, bool $pass_route = false): void
+    public function _post(string $pattern, callable $callback, bool $pass_route = false, string $route_alias = ''): void
     {
-        $this->router()->map('POST ' . $pattern, $callback, $pass_route);
+        $this->router()->map('POST ' . $pattern, $callback, $pass_route, $route_alias);
```
n0nag0n, line 558: Good catch :)

flight/Engine.php
```diff
@@ -618,26 +613,22 @@ public function _halt(int $code = 200, string $message = ''): void
         }
     }
 
-    /**
-     * Sends an HTTP 404 response when a URL is not found.
-     */
+    /** Sends an HTTP 404 response when a URL is not found. */
     public function _notFound(): void
     {
+        $output = '<h1>404 Not Found</h1><h3>The page you have requested could not be found.</h3>';
```
n0nag0n, line 619: if you wanted to be mega fancy you could do
```php
$output = <<<HTML
    <h1>404 Not Found</h1><h3>The page you have requested could not be found.</h3>
    HTML;
```
and then you'd have syntax highlighting ;)

flight/Flight.php
```diff
@@ -104,56 +101,53 @@ private function __clone()
...
+     * @param array<int, mixed>  $params   Class constructor params
+     * @param ?Closure(T $instance): void $callback Perform actions with the instance
+     *
+     * @template T of object
      */
-    public static function register($name, $class, $params = [], $callback = null)
+    public static function register($name, $class, $params = [], $callback = null): void
     {
-        static::__callStatic('register', func_get_args());
+        static::__callStatic('register', [$name, $class, $params, $callback]);
```
n0nag0n, line 123: I've wondered if these 2 methods even need to exist.....cause __callStatic will just call them anyways.....seems weird.

flight/core/Dispatcher.php
```diff
@@ -18,45 +20,53 @@
...
      */
     public function run(string $name, array $params = [])
     {
         $output = '';
 
         // Run pre-filters
-        if (!empty($this->filters[$name]['before'])) {
+        $thereAreBeforeFilters = !empty($this->filters[$name]['before']);
+
+        if ($thereAreBeforeFilters) {
```
n0nag0n, line 54: Code as documentation, I like it!

flight/core/Dispatcher.php
```diff
@@ -66,20 +76,28 @@ public function run(string $name, array $params = [])
...
-     * @param callable $callback Callback function
+     * @param string $name Event name
+     * @param Closure(): (void|mixed) $callback Callback function
+     *
+     * @return $this
      */
-    public function set(string $name, callable $callback): void
+    public function set(string $name, callable $callback): self
     {
+        if ($this->get($name) !== null) {
```
n0nag0n, line 87: So I see what you're doing here, but this may actually cause other peoples older code to break. They might actually have bad code in place and this will force an error to be thrown and break their app, which is breaking backwards compatibility

flight/core/Dispatcher.php
```diff
@@ -66,20 +76,28 @@ public function run(string $name, array $params = [])
...
+            trigger_error("Event '$name' has been overriden!", E_USER_NOTICE);
+        }
+
         $this->events[$name] = $callback;
+
+        return $this;
```
n0nag0n, line 93: But this should be fine

flight/core/Dispatcher.php
```diff
@@ -105,42 +123,59 @@ public function has(string $name): bool
...
+     * @param string $name Event name
+     * @param 'before'|'after' $type Filter type
+     * @param Closure(array<int, mixed> &$params, string &$output): (void|false) $callback
+     *
+     * @return $this
      */
-    public function hook(string $name, string $type, callable $callback): void
+    public function hook(string $name, string $type, callable $callback): self
     {
+        if (!in_array($type, self::FILTER_TYPES, true)) {
```
n0nag0n, line 149: If you want to take this one step further `in_array($type, self::FILTER_TYPES, true) === false` is actually slightly more performent because `!` is a falsey evaluation and it has to evaluate multiple potential false values (false, 0, '', etc)

flight/core/Dispatcher.php
```diff
@@ -105,42 +123,59 @@ public function has(string $name): bool
...
+            $noticeMessage = "Invalid filter type '$type', use " . join('|', self::FILTER_TYPES);
+
+            trigger_error($noticeMessage, E_USER_NOTICE);
```
n0nag0n, line 152: This is another one of those things. I don't know if someone set up `'afterEmail'` as a hook in an old app, and they used it in a weird way outside of the framework, (which you can use cause a lot of these Dispatcher methods are public). I would say you'd have to revert this.

flight/core/Dispatcher.php
```diff
@@ -149,62 +184,69 @@ public function filter(array $filters, array &$params, &$output): void
...
+     * @throws TypeError For unexistent class name.
      */
-    public static function invokeMethod($func, array &$params = [])
+    public static function invokeMethod(array $func, array &$params = [])
     {
         [$class, $method] = $func;
 
-        $instance = \is_object($class);
+        if (is_string($class) && class_exists($class)) {
+            $class = new $class();
```
n0nag0n, line 257: Honestly this change makes me happy. This is something on my mental todo list! I was going to take it a step further where you could if you wanted to just do something like `Flight::route('/url', 'Controller->method');` but this is good for now

flight/core/Dispatcher.php
n0nag0n: After going through the unit test, I still don't know if this is something people could have overwritten in the past where they setup their custom hooks. I just don't know what I don't know :(

n0nag0n (changes requested):

flight/core/Dispatcher.php
n0nag0n: Oh one thing you will need to change is that this need to inject values into the constructor. Likely Flight::app(), or $app if that's somehow available from the Dispatcher or something similar

n0nag0n (changes requested):

flight/Engine.php
fadrian06: Before finishing the PR, I ran phpcs and it only reported the use of _method() in Engine

flight/Engine.php
fadrian06: The cyclomatic complexity is still 2, there is still an implicit if else, so testing both branches would give 100% coverage

.editorconfig
fadrian06: The thing is that the readme examples look better with 2 spaces, because it makes it easier to read on mobile. the 2 spaces only apply to .md, the rest follow 4 spaces

flight/Engine.php
fadrian06: I extracted a variable just to reduce the line size and phpcs will stop complaining. But I also like heredoc for these things

flight/Flight.php
fadrian06: I did it like this because of a phpcs rule about unused parameters

flight/core/Dispatcher.php
fadrian06: I'll take a look at the scenarios that could throw an exception. Could you provide a use case that would fail now due to an exception and not before? Could you provide a use case that would fail now due to an exception and not before?

flight/core/Dispatcher.php
fadrian06: I'll fix it :)

flight/core/Dispatcher.php
fadrian06: Using an invalid filter previously seemed to work but when it came to running the filters, it ignored them, I thought to throw an exception for invalid filters It would be very beast, so I thought of issuing a notice that does not stop the execution of the code

fadrian06: > After going through the unit test, I still don't know if this is something people could have overwritten in the past where they setup their custom hooks. I just don't know what I don't know :(

Ah, I understand what you mean, by inheriting from the class you can increase the number of filters. Okay, I'll fix that.

fadrian06: Although coming back to think, it doesn't make much sense to add more filters to an event, in the execution cycle There are only 2 moments of the method, before and after, and these can already be chained

flight/core/Dispatcher.php
fadrian06: Precisely for this reason I wanted to add a DI container, which is just used for dependencies of a constructor.

flight/core/Dispatcher.php
n0nag0n: I guess I'm thinking that in someone's code they might have something like `Dispatcher->set('someMethod')` and then in a different part of their code, like a controller, they might override that method with the same name to do a different behavior.

I'm not agreeing that that is a good way to go, but I'm just saying it's possible. Going through all the issues and cleaning them up showed that some people have really bizarre work arounds for things like pseudo-middleware and whatnot.

flight/core/Dispatcher.php
n0nag0n: > I remember some people adding things like `Flight::route('*', function() {});` to do middleware as work arounds, .....

```php
<?php

// require 'flight/Flight.php';
require 'flight/autoload.php';

Flight::route('*', function () {
    echo 'HELLO FROM MIDDLEWARE <br />';

    return true;
});

Flight::route('/', function () {
    echo 'hello world!';
});

Flight::start();
```

```bash
php -S localhost:80
> [Sun Feb  4 18:05:18 2024] PHP 7.4.33 Development Server (http://localhost:80) started
```

```bash
curl http://localhost
> HELLO FROM MIDDLEWARE <br />hello world!
```

krmu (changes requested):
Fatal error comes in. Knock, knock.
` [200]: GET / - Uncaught Error: Class "flight\core\Dispatcher" not found in /FlightPHP/flight/Flight.php:141
Stack trace:
#0 /FlightPHP/index.php(5): Flight::__callStatic('route', Array)
#1 {main}
  thrown in /FlightPHP/flight/Flight.php on line 141`

fadrian06:
I guess there are no tests for that case.

flight/core/Dispatcher.php
fadrian06: Well, to start, it's `(new Dispatcher)->set('someMethod')` instead of `Dispatcher::set('someMethod')`.

So the only case when a method'll be overriden an therefore thrown a E_NOTICE is when the controller use the same object that it's used in other app parts

fadrian06: I prefer to throw an E_NOTICE instead of an Exception to not make a breaking change

flight/core/Dispatcher.php
fadrian06: [`in_array(mixed $needle, array $haystack, bool $strict = false): bool`](https://www.php.net/manual/es/function.in-array.php).

That's because I can use `!` safety that `in_array` always returns a `bool`.

fadrian06:
Meanwhile Flight give [PSR-11](https://www.php-fig.org/psr/psr-11/) support, cannot call an instance method statically when class require constructor parameters

krmu (approved):
Everything works as expected.
But getting now notice, should be so?
`Notice: Event 'error' has been overriden! `
`Notice: Event 'notFound' has been overriden! `
Otherwise everything works.

fadrian06:
With the release of Notice and the features added to Dispatcher, flight could move to 3.4.0

n0nag0n (approved):

PR #539, opened 2024-02-06
fixed bug with grouped routes

n0nag0n:
This fixes an issue where a grouped route such as the below fails.

```php
Flight::group('/api', function($router) {
    Flight::route('', function() { echo 'Hi'; });
    // or
    Flight::route('GET|POST', function() { echo 'Hi'; });
});
```

If you navigate to `/api` this route should still work.

### DIFF
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -64,7 +64,6 @@ use flight\net\Route;
  * @method void etag(string $id, ('strong'|'weak') $type = 'strong') Handles ETag HTTP caching.
  * @method void lastModified(int $time) Handles last modified HTTP caching.
  */
-// phpcs:ignoreFile PSR2.Methods.MethodDeclaration.Underscore
 class Engine
 {
     /** @var array<string, mixed> Stored variables. */
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -64,6 +64,7 @@ use flight\net\Route;
  * @method void etag(string $id, ('strong'|'weak') $type = 'strong') Handles ETag HTTP caching.
  * @method void lastModified(int $time) Handles last modified HTTP caching.
  */
+// phpcs:ignoreFile PSR2.Methods.MethodDeclaration.Underscore
 class Engine
 {
     /** @var array<string, mixed> Stored variables. */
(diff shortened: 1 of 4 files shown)


krmu (approved):
From my tests, everything works.

flight/Engine.php
```diff
@@ -64,6 +64,7 @@
  * @method void etag(string $id, ('strong'|'weak') $type = 'strong') Handles ETag HTTP caching.
  * @method void lastModified(int $time) Handles last modified HTTP caching.
  */
+// phpcs:ignoreFile PSR2.Methods.MethodDeclaration.Underscore
```
fadrian06: If we'll have phpcs to ignore rules? what's the sense of that?

flight/core/Dispatcher.php
```diff
@@ -78,7 +78,7 @@ public function run(string $name, array $params = [])
      * Assigns a callback to an event.
      *
      * @param string $name Event name
-     * @param Closure(): (void|mixed) $callback Callback function
+     * @param Closure(): (void|mixed)|callable $callback Callback function
```
fadrian06: callable and Closure(): (void|mixed) are same... I use Closure because it allows any valid callable and LSP-intelephense can detect that in `$result = $myParamClosure()`, $result'll be `@var void|mixed`...

And with simple `callable` doc typehint, $result'll be detected as `@var mixed`

fadrian06 (commented):
Only remove @phpstan-ignore### and
Closure(): (...) | callable to simply Closure(): (...) because the other is redundant.

flight/Engine.php
n0nag0n: All I did was run composer phpcs and all those private methods that start with _ failed. This just ignores those errors for this file.

flight/core/Dispatcher.php
n0nag0n: So I literally did nothing but running phpstan made this rule fail. So I added the little callable line and then it passed.

fadrian06:
## Incomplete typing (Closure() ... | callable)
image

## Complete typing (Closure() ...)
image

PR #543, opened 2024-02-09
Doesn't throw NOTICE for notFound and error overrides

fadrian06:
## Now Engine::map is chainneable.

```php
$engine = new Engine;

$engine->map(...)
  ->map(...)
  ->map(...);
```

### DIFF
--- a/flight/core/Dispatcher.php
+++ b/flight/core/Dispatcher.php
@@ -116,7 +116,9 @@ class Dispatcher
      */
     public function set(string $name, callable $callback): self
     {
-        if ($this->get($name) !== null) {
+        static $allowedMethods = ['notFound', 'error'];
+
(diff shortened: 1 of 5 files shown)


flight/core/Dispatcher.php
```diff
@@ -116,7 +116,9 @@ protected function runPostFilters(string $eventName, &$output)
      */
     public function set(string $name, callable $callback): self
     {
-        if ($this->get($name) !== null) {
+        static $allowedMethods = ['notFound', 'error'];
```
n0nag0n: Sorry this won't work. We have to remove the whole error message because every single mapped method will throw an error.

fadrian06: okay

PR #546, opened 2024-02-18
Removed broken image in README

fadrian06:


n0nag0n:
Looks like they are having server issues. https://hits.dwyl.com/flightphp/core

```
Internal Server Error
```

Let's see if it resolves itself in a couple days and if not, we'll remove it.

n0nag0n (approved):
It's still not better....we can remove it.

PR #553, opened 2024-03-13
fix: handle encoded slashes for url parameters - fixes #552

### LINKED ISSUE #552: Handling of URLs with encoded slashes (/)
Currently it is not possible to have URL parameters which contain (encoded) slashes.

E.g. for the route `/redirect/@id`  the request url `/redirect/before%2Fafter` should result in `$id = 'before/after'`.

AFAIK this is the standard behavior in other implementations.

daniel-sc:
Caveat: technically, this introduces a backwards incompatible change, for instances where in the request URL (regular) slashes are encoded. I'd guess, that this is not relevant in regular setups and even though this was gracefully handled before, this change is neccessary to assure path params can contain encoded slashes.

### DIFF
--- a/flight/net/Route.php
+++ b/flight/net/Route.php
@@ -97,7 +97,7 @@ class Route
     /**
      * Checks if a URL matches the route pattern. Also parses named parameters in the URL.
      *
-     * @param string $url            Requested URL
+     * @param string $url            Requested URL (original format, not URL decoded)
      * @param bool   $case_sensitive Case sensitive matching
      *
      * @return bool Match status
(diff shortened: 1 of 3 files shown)


n0nag0n:
I can take a look at it in hopefully tonight or tomorrow. Thanks for the contribution and for testing it :)

n0nag0n:
I added some additional testing, but did get this merged in. Thanks so much! https://github.com/flightphp/core/commit/765887d388aff855cf4ffad8fef2df8ef3859025

PR #555, opened 2024-03-16
Also register the HEAD method alongside GET

BelleNottelling:
Hi all! 

First off, thanks for Flight it's honestly a great little framework and It's been a treat to work with.

This is a very small patch that causes Flight to automatically register the HEAD method alongside GET if the user didn't do so manually. Per [the RFC](https://www.rfc-editor.org/rfc/rfc9110.html#HEAD), a server should respond identically to a HEAD request as a GET request except with only the headers.

The web-server in-front of PHP should ensure only the headers are sent back to the client, meaning the only aspect Flight needs to handle is just ensuring a HEAD request is handled the same as a GET request.

In terms of real-world use-case, I know that some uptime monitors (specifically updown.io) will try to use HEAD requests to save a little bit of bandwidth which would then fallback to GET requests when used with Flight.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
     },
     "scripts": {
         "test": "phpunit",
-        "test-coverage": "rm clover.xml && XDEBUG_MODE=coverage vendor/bin/phpunit --coverage-html=coverage --coverage-clover=clover.xml && vendor/bin/coverage-check clover.xml 100",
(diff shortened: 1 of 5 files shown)


n0nag0n (approved):
I had to go in and add proper testing, and to make sure the body is not output on a head request. Thanks for contributing to make Flight better! I'm glad it's fun for you to work with! This will be merged in now and appended to the next release. Also, tell your friends about Flight :laughing:

BelleNottelling:
Cheers and thanks!

PR #566, opened 2024-03-30
Added ability to load classes with _ in them

n0nag0n:
This allows you to load a class in your codebase such as `Pascal_Snake_Case` and it will still work. You must set it with `Loader::setV2ClassLoading(false);`

### DIFF
--- a/flight/util/ReturnTypeWillChange.php
+++ b/flight/util/ReturnTypeWillChange.php
@@ -1,9 +1,0 @@
-<?php
-
-declare(strict_types=1);
-
-// This file is only here so that the PHP8 attribute for doesn't throw an error in files
-// phpcs:ignoreFile PSR1.Classes.ClassDeclaration.MissingNamespace
-class ReturnTypeWillChange
-{
-}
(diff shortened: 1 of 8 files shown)


flight/util/ReturnTypeWillChange.php
n0nag0n, line 1: This actually caused php8.1+ to fail because it thought the ReturnTypeWillChange class was an attribute class....so it failed.

tests/EngineTest.php
```diff
@@ -751,8 +751,14 @@ public function testContainerDicePdoWrapperTestBadParams() {
         $engine->route('/container', Container::class.'->testThePdoWrapper');
         $engine->request()->url = '/container';
 
-        $this->expectException(ErrorException::class);
-        $this->expectExceptionMessageMatches("/Passing null to parameter/");
+        // php 7.4 will throw a PDO exception, but php 8 will throw an ErrorException
+        if(version_compare(PHP_VERSION, '8.0.0', '<')) {
```
n0nag0n, line 755: This is highly unusual, but the behavior in PDO actually changed inbetween 7.4 and 8.0 to throw different error messages.

PR #567, opened 2024-04-01
Improvements to Dispatcher class

fadrian06:
# Tired of declaring after filters with the first parameter without doing anything?

### **DELETE IT :D, nothing will be broken.**

Dispatcher now supports after filters with the only useful parameter.

```php
Flight::after('myMethod', function (&$params, &$output) {
  // $params Even if it is used it does not affect, but it had to be declared yes or yes even if it does nothing.
});
```

## Now...

```php
Flight::after('myMethod', function (&$output) {
  // goodbye useless $params 
});
```

> **IMPORTANT:** both ways work exactly the same. Only one is shorter and cleaner.

To even make it even cleaner, you could change the filter to a PHP 7.4 arrow function.

```php
Flight::after('myMethod', fn (&$output) => doMagic($output));
```

**Just make sure that what you do in the arrow function doesn't return `false`**, unless you want to make other after filters for the same mapped method not run.


.gitignore
```diff
@@ -6,3 +6,4 @@ composer.lock
 coverage/
 *.sublime*
 clover.xml
+phpcs.xml
```
n0nag0n, line 9: I guess I don't understand the benefit to a phpcs.xml.dist over a phpcs.xml. It just looks like it's now more complicated with post-install-cmds to copy it over and gitignore is ignoring it. Seems like a lot of work to just.....do the same thing?

flight/core/Dispatcher.php
```diff
@@ -23,41 +24,44 @@ class Dispatcher
...
+     * Dependency injection container.
      */
     public function setContainerHandler($containerHandler): void
     {
-        $this->containerHandler = $containerHandler;
+        if (
+            $containerHandler instanceof ContainerInterface
+            || is_callable($containerHandler)
+        ) {
+            $this->containerHandler = $containerHandler;
```
n0nag0n, line 68: So this chunk of code doesn't bring any additional value, if anything it will confuse people when they set something incorrectly and no error is thrown, but the value didn't save, they won't know why til they start to debug. At least with the previous way it would get set and an error would be thrown letting them know they injected the wrong object. If anything, you need to add an error message or throw an exception to let people know that they didn't set the right variable.

flight/core/Dispatcher.php
```diff
@@ -310,69 +326,59 @@ public function invokeMethod(array $func, array &$params = [])
...
      * @throws TypeError For nonexistent class name.
-     * @throws InvalidArgumentException If the constructor requires parameters
+     * @throws InvalidArgumentException If the constructor requires parameters.
      * @version 3.7.0
      */
     public function invokeCallable($func, array &$params = [])
     {
         // If this is a directly callable function, call it
-        if (is_array($func) === false) {
+        if (!is_array($func)) {
```
n0nag0n: I mean it does the same thing, but I don't know why? === false is actually more performant and explicit.

flight/core/Dispatcher.php
```diff
@@ -310,69 +326,59 @@ public function invokeMethod(array $func, array &$params = [])
...
-        ) {
-            $containerHandler = $this->containerHandler;
-            $resolvedClass = $this->resolveContainerClass($containerHandler, $class, $params);
-            if ($resolvedClass !== null) {
+        $mustUseTheContainer = $this->containerHandler && (
+            (is_object($class) && !str_starts_with(get_class($class), 'flight\\'))
+            || is_string($class)
+        );
+
+        if ($mustUseTheContainer) {
```
n0nag0n: I'd add === true to this. I'm sure all the unit tests pass, but this chunk of code was messy. If this still works, kudos for cleaning it up!

flight/core/Dispatcher.php
```diff
@@ -310,69 +326,59 @@ public function invokeMethod(array $func, array &$params = [])
...
-                    strpos(get_class($class), 'flight\\') === false
-                ) ||
-                is_string($class) === true
-            )
-        ) {
-            $containerHandler = $this->containerHandler;
-            $resolvedClass = $this->resolveContainerClass($containerHandler, $class, $params);
-            if ($resolvedClass !== null) {
+        $mustUseTheContainer = $this->containerHandler && (
+            (is_object($class) && !str_starts_with(get_class($class), 'flight\\'))
```
n0nag0n: You'd need to add `=== true` here in a couple cases. We cannot use `str_starts_with` as that is PHP8 functionality and I don't want polyfill dependencies.

composer.json
```diff
@@ -24,7 +24,9 @@
     ],
     "require": {
         "php": "^7.4|^8.0|^8.1|^8.2|^8.3",
-        "ext-json": "*"
+        "ext-json": "*",
+        "psr/container": "^2.0",
+        "symfony/polyfill-php80": "^1.29"
```
n0nag0n: Sorry, no polyfills

composer.json
```diff
@@ -24,7 +24,9 @@
     ],
     "require": {
         "php": "^7.4|^8.0|^8.1|^8.2|^8.3",
-        "ext-json": "*"
+        "ext-json": "*",
+        "psr/container": "^2.0",
```
n0nag0n: I know you have this in here for one single comment. I don't know if adding a dependency is worth a comment. Given the stuff that's happened with xz, I'm scrutinizing every dependency I can.

flight/core/Dispatcher.php
```diff
@@ -381,84 +387,77 @@ protected function verifyValidFunction($callback): void
...
-     *
-     * @return void
      */
     protected function verifyValidClassCallable($class, $method, $resolvedClass): void
     {
-        $final_exception = null;
+        $exception = null;
 
         // Final check to make sure it's actually a class and a method, or throw an error
-        if (is_object($class) === false && class_exists($class) === false) {
```
n0nag0n, line 397: Why did your code editor get rid of all these? Could you configure it to not do that please?

flight/util/ReturnTypeWillChange.php
```diff
@@ -0,0 +1,8 @@
+<?php
```
n0nag0n, line 1: I removed this on purpose because on PHP8, PHPStan will thrown errors because it complains this class shouldn't exist.

n0nag0n (changes requested):
I see you did some improvements which if they work out will help the code be cleaner which I appreciate! Your code editor is messing up a few things that need to be corrected though. And the only other thing is that I don't want to introduce dependencies into Flight, that's part of the selling point.

flight/core/Dispatcher.php
fadrian06: The flight documentation says that the method must be passed either an instance that implements psr11 or a valid callable, the docblock says it accepts a psr11 instance or a valid callable . Why would the user pass something invalid?

flight/core/Dispatcher.php
fadrian06: I also read code in English, yes I already know that the symbol ! is not, then it is read as a sentence: If not is an array...

flight/core/Dispatcher.php
fadrian06: Plenary session 2024 with 5MB/s bandwidth, does 5kb extra matter?

fadrian06: Plenary session 2024 with 5MB/s bandwidth, does 5kb extra matter?

composer.json
fadrian06: You keep this then, I'm out

n0nag0n:
@fadrian06 Why'd you close the PR? It's a good PR just needs a few tweaks

.gitignore
fadrian06: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-default-configuration-file

fadrian06: The dist defines the basic rules that every contributor will have when running the linter. It is optional to define more rules so that it detects more things (as in my case), (as I don't want all the super strict configurations to be available to everyone) then the dist will have the simple ones and the one that doesn't have dist and git ignores, will have your custom configurations

flight/core/Dispatcher.php
fadrian06: then allow me to generate an InvalidArgumentException?

flight/core/Dispatcher.php
```diff
@@ -4,10 +4,11 @@
 
 namespace flight\core;
 
-use Closure;
 use Exception;
 use flight\Engine;
 use InvalidArgumentException;
+use Psr\Container\ContainerInterface;
```
n0nag0n, line 10: And this needs to be removed right?

flight/core/Dispatcher.php
```diff
@@ -23,41 +24,54 @@ class Dispatcher
...
-     * @var array<string, array<'before'|'after', array<int, Closure(array<int, mixed> &$params, mixed &$output): (void|false)>>>
+     * @var array<string, array<'before'|'after', array<int, callable(array<int, mixed> &$params, mixed &$output): (void|false)>>>
      */
     protected array $filters = [];
 
     /**
      * This is a container for the dependency injection.
      *
-     * @var callable|object|null
+     * @var null|ContainerInterface|(callable(string $classString, array<int, mixed> $params): (null|object))
```
n0nag0n, line 48: Do you still need to correct this comment?

flight/core/Dispatcher.php
```diff
@@ -23,41 +24,54 @@ class Dispatcher
 {
     public const FILTER_BEFORE = 'before';
     public const FILTER_AFTER = 'after';
-    private const FILTER_TYPES = [self::FILTER_BEFORE, self::FILTER_AFTER];
 
-    /** @var mixed $containerException Exception message if thrown by setting the container as a callable method */
-    protected $containerException = null;
+    /** Exception message if thrown by setting the container as a callable method. */
+    protected ?Exception $containerException = null;
```
n0nag0n: So I.....think this will work? This is an exception handled by the container exception. If they don't extend Exception for some reason, then this will fail. I just don't know what I don't know about how PHP-DI, Dice, Aura, et'c handle their exceptions.

flight/core/Dispatcher.php
fadrian06: don't forget, it doesn't affect runtime at all, within flight it works because Dice also installs the interface, flight clients that don't use the interface are not affected

flight/core/Dispatcher.php
```diff
@@ -188,27 +202,38 @@ public function clear(?string $name = null): void
...
+     * @param 'before'|'after' $type Filter type.
+     * @param callable(array<int, mixed> &$params, mixed &$output): (void|false)|callable(mixed &$output): (void|false) $callback
      *
      * @return $this
      */
     public function hook(string $name, string $type, callable $callback): self
     {
-        if (!in_array($type, self::FILTER_TYPES, true)) {
-            $noticeMessage = "Invalid filter type '$type', use " . join('|', self::FILTER_TYPES);
+        static $filterTypes = [self::FILTER_BEFORE, self::FILTER_AFTER];
```
n0nag0n, line 220: It doesn't need to be static does it? Since it's just referring to static types. The other places in the framework that use this are those places trying to avoid using $this->someVar to hold data.

flight/core/Dispatcher.php
fadrian06: yeah, before it was callable(... array) and phpstan was complaining that I told him array of what. Well, since they are arguments of a constructor, well, mixed

flight/core/Dispatcher.php
fadrian06: What is above exception is the throwable interface, unless the DI containers return Error instead of Exception, the property will accept the reassignment correctly

fadrian06: All PSR/11 must either throw the exception required by the interface, the only thing that can fail is that the callable provided as the handler throws an Error and not an Exception

fadrian06: All PSR/11 must either throw the exception required by the interface, the only thing that can fail is that the callable provided as the handler throws an Error and not an Exception

fadrian06: I already changed it and uploaded it

flight/core/Dispatcher.php
fadrian06: Since the only place that used the constant was that function, I moved it to the function, the fact that it is static in the function indicates that it is only initialized once, this is for performance and micro-optimization

https://www.php.net/manual/en/language.variables.scope.php

look for static variables

n0nag0n (approved):

PR #571, opened 2024-04-09
added ability to convert the response body

n0nag0n:
This will allow you to minify, gzip, whatever you'd like to do with the response body.

```php
Flight::response()->addResponseBodyCallback(function($body) {
    return gzencode($body, 9);
});
```

PR #572, opened 2024-04-11
Fix content length when using response body callbacks

BelleNottelling:
#571 introduced an issue where the response body could be modified **after** Flight calculates & sends the content length to the browser. Since the response body callback modifies the body, the content length is no longer correct.

In my testing things seemed to still work okay locally with the PHP dev server, but once I moved to an [actual server](https://api.fossbilling.org/) I noticed the browser would sit and continue to try to load after the body has already sent as the response body was compressed after the content length was sent.

Switching the order so the callback is called and then headers are sent fixes this

### DIFF
--- a/flight/net/Response.php
+++ b/flight/net/Response.php
@@ -341,6 +341,15 @@ class Response
             );
         }
 
+        if ($this->content_length === true) {
+            // Send content length
+            $length = $this->getContentLength();
+
+            if ($length > 0) {
+                $this->setHeader('Content-Length', (string) $length);
+            }
+        }
+
         // Send other headers
         foreach ($this->headers as $field => $value) {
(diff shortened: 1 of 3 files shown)


n0nag0n:
Want to give my changes a go real quick to make sure I didn't mess up anything on your end?

BelleNottelling:
> Want to give my changes a go real quick to make sure I didn't mess up anything on your end?

Still works as expected on my end, thanks for checking

n0nag0n (approved):

n0nag0n:
Merged and released as v3.8.1. Thanks for the assist!

BelleNottelling:
Thanks for getting this merged and releasing the fix!

PR #585, opened 2024-05-02
Allowed for middlewares to be created with container

n0nag0n:
This PR allows for middlewares to be called with just class names `Middleware::class` for use with and without a container handler.

### DIFF
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -382,64 +382,82 @@ class Engine
      * Processes each routes middleware.
      *
      * @param Route $route The route to process the middleware for.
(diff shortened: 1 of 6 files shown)


fadrian06 (commented):
Thank you for your work on improving the integration with the container.

Just a suggestion, I had a hard time reading the diff between the change from snake to camelCase, so it's okay to separate it into 2 commits that are free :D

n0nag0n:
Ahh yeah.....I could have, but it was just bugging the crap out of me so I did it in the one commit haha!

PR #586, opened 2024-05-05
Changed it so runway commands are run from any repo

n0nag0n:
Runway is now just a shell, and each of the commands lives in a prospective repo that's defined in the paths for the runway repo itself.

So there will be a few commands in this core repo, there will be some in the active-record repo if you choose to have that repo in your project, and based on the dependencies you add, if they have specific commands for runway (usually saved in the `src/commands/` directory), you can now use them automatically! :partying_face:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage/
 *.sublime*
 clover.xml
 phpcs.xml
+.runway-config.json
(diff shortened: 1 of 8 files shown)


fadrian06:
woa, more dependencies of the dependency..fine :D

n0nag0n:
So this is only a dev dependency. When you install flightphp/core, it does add the dependency. I had to add it as a dev dependency so it could be properly unit tested.

PR #591, opened 2024-05-17
Added ability to throw a method not found instead of 404

n0nag0n:
If you are routing a URL and it finds the URL, but not the method, this allows a 405 Method Not Found error to be thrown instead of the expected 404 Not Found.

### DIFF
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -388,15 +388,14 @@ class Engine
     {
         $atLeastOneMiddlewareFailed = false;
 
-		// Process things normally for before, and then in reverse order for after.
-        $middlewares = $eventName === Dispatcher::FILTER_BEFORE 
-			? $route->middleware 
-			: array_reverse($route->middleware);
+        // Process things normally for before, and then in reverse order for after.
+        $middlewares = $eventName === Dispatcher::FILTER_BEFORE
+            ? $route->middleware
+            : array_reverse($route->middleware);
         $params = $route->params;
 
         foreach ($middlewares as $middleware) {
-
-			// Assume that nothing is going to be executed for the middleware.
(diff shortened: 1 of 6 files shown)


flight/Engine.php
```diff
@@ -388,15 +388,14 @@ protected function processMiddleware(Route $route, string $eventName): bool
...
 
-		// Process things normally for before, and then in reverse order for after.
-        $middlewares = $eventName === Dispatcher::FILTER_BEFORE 
-			? $route->middleware 
-			: array_reverse($route->middleware);
+        // Process things normally for before, and then in reverse order for after.
+        $middlewares = $eventName === Dispatcher::FILTER_BEFORE
+            ? $route->middleware
+            : array_reverse($route->middleware);
         $params = $route->params;
```
n0nag0n, line 395: This will look better if you remove the whitespace setting in the CR

flight/Engine.php
```diff
@@ -579,7 +578,13 @@ public function _start(): void
         if ($failedMiddlewareCheck === true) {
             $this->halt(403, 'Forbidden', empty(getenv('PHPUNIT_TEST')));
         } elseif ($dispatched === false) {
-            $this->notFound();
+            // Get the previous route and check if the method failed, but the URL was good.
+            $lastRouteExecuted = $router->executedRoute;
+            if ($lastRouteExecuted !== null && $lastRouteExecuted->matchUrl($request->url) === true && $lastRouteExecuted->matchMethod($request->method) === false) {
+                $this->halt(405, 'Method Not Allowed', empty(getenv('PHPUNIT_TEST')));
```
luka97zr, line 584: Sweet, I like this, so helpful for debugging!

luka97zr (approved):
Thanks for making these changes that fast, so cool!

PR #597, opened 2024-06-07
fix: make the jsonHalt a static function as the user guide indicates

Billtec:


n0nag0n:
Yup, looks like a comment was missed, thanks!

PR #605, opened 2024-08-27
add case for Macos in ResponseTest::testResponseBodyCallbackGzip

lubiana:
the created gzip seems to be different on macos, so i addeed the correct version


n0nag0n (approved):

n0nag0n:
Thanks so much!

PR #606, opened 2024-08-27
add unit-test workflow

lubiana:
this adds a github workflow that runs the unit tests against all php versions from 7.4 up to 8.4 on all pull requests

### DIFF
--- a/tests/commands/ControllerCommandTest.php
+++ b/tests/commands/ControllerCommandTest.php
@@ -68,6 +68,8 @@ class ControllerCommandTest extends TestCase
 
     public function testCreateController()
     {
+
+        $this->markTestIncomplete('does not work on php > 8.0');
         $app = $this->newApp('test', '0.0.1');
         $app->add(new ControllerCommand(['app_root' => 'tests/commands/']));
         $app->handle(['runway', 'make:controller', 'Test']);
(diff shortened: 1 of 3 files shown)


lubiana:
we might think about disabling the test-run for 8.4 for now, as the problems come from external dependencies

tests/commands/ControllerCommandTest.php
```diff
@@ -68,6 +68,8 @@ public function testControllerAlreadyExists()
 
     public function testCreateController()
     {
+
+        $this->markTestIncomplete('does not work on php > 8.0');
```
n0nag0n, line 72: I wonder why this one doesn't work. I've been able to run php8.3 unit tests on this without any problem...I wonder wonder why it gave you problems.

n0nag0n: I even went to my core code, removed my vendor dir, installed everything with php8.2 and my tests still pass.

lubiana: i will check again

lubiana: actually i am not sure if it is really a php8 problem
 but steps to reproduce should be to complety remove the vendor dir and composer.lock, make sure the dependency for nette/php-generator is not in composer.json, then do a composer install and run the test again

n0nag0n: aha, have to also delete the composer.lock directory. Got it!

n0nag0n (approved):

PR #618, opened 2025-01-14
Fixed multiple trailing slashes issues. 

n0nag0n:
Also fixed server test for 405 and 404

### DIFF
--- a/flight/net/Route.php
+++ b/flight/net/Route.php
@@ -98,16 +98,23 @@ class Route
      * Checks if a URL matches the route pattern. Also parses named parameters in the URL.
      *
      * @param string $url            Requested URL (original format, not URL decoded)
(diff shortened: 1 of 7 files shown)


flight/net/Route.php
```diff
@@ -98,17 +98,24 @@ public function __construct(string $pattern, $callback, array $methods, bool $pa
...
             return true;
         }
 
+        // if the last character of the incoming url is a slash, only allow one trailing slash, not multiple
+        if (substr($url, -2) === '//') {
+            // remove all trailing slashes, and then add one back.
+            $url = rtrim($url, '/') . '/';
+        }
+
+
```
n0nag0n, line 118: This is the code above that actually fixes the problem.

PR #620, opened 2025-02-03
Added PUT, PATCH, DELETE methods for data

KnifeLemon:


n0nag0n:
Hey there, thanks for the contribution! I will try to get to this in the next couple of days (been slammed lately). Did you happen to run `composer test-coverage` and was the coverage 100%?

n0nag0n:
So I went ahead and fixed the unit test errors. I had to correct some of the logic as I think some of what you wanted would have caused problems for certain types of requests. Can you look over the final PR and let me know if I solved the problem in the way you were hoping? I'm assuming that you were shooting for form encoded values and not JSON request correct?

KnifeLemon:
> So I went ahead and fixed the unit test errors. I had to correct some of the logic as I think some of what you wanted would have caused problems for certain types of requests. Can you look over the final PR and let me know if I solved the problem in the way you were hoping? I'm assuming that you were shooting for form encoded values and not JSON request correct?

Yes, the way you wrote it is correct. It is correct to handle form encoded values, not json format.

n0nag0n:
Thanks so much!

KnifeLemon:
> Thanks so much!

Can it also include multipart data?

Thank you for your quick response and work ! 👍

PR #623, opened 2025-02-14
Simplify Flight and faster performance

joanhey:
Null coalescing operator work from PHP 7.0

I don't know if is it necessary the `setEngine()` because is public ?
I think that is for use an extended Engine. But before never will work if changed before any call, because don't change the `$initialized`.

### DIFF
--- a/flight/Flight.php
+++ b/flight/Flight.php
@@ -87,9 +87,6 @@ class Flight
     /** Framework engine. */
     private static Engine $engine;
 
-    /** Whether or not the app has been initialized. */
-    private static bool $initialized = false;
-
     /**
      * Don't allow object instantiation
      *
@@ -127,14 +124,7 @@ class Flight
     /** @return Engine Application instance */
     public static function app(): Engine
     {
-        if (!self::$initialized) {
-            require_once __DIR__ . '/autoload.php';
-
-            self::setEngine(new Engine());
-            self::$initialized = true;
-        }
-
-        return self::$engine;
+        return self::$engine ?? self::$engine = new Engine();
     }
 
     /**
@@ -142,8 +132,8 @@ class Flight
      *
      * @param Engine $engine Vroom vroom!
      */
-    public static function setEngine(Engine $engine): void
+    public static function setEngine(Engine $engine): Engine
     {
-        self::$engine = $engine;
+        return self::$engine = $engine;
     }
 }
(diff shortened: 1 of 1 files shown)


joanhey:
I see that you use it in the tests:

```php
protected function setUp(): void
    {
        $_SERVER = [];
        $_REQUEST = [];
        Flight::init();
        Flight::setEngine(new Engine());
        Flight::set('flight.views.path', __DIR__ . '/views');
    }

```

But when you call the `Flight::init()`, automatically set the engine with `new Engine()` if not set.
I don't understand why later call `Flight::setEngine(new Engine());` and also invalidate the before `init()`.

It'll work the same with:
```php
protected function setUp(): void
    {
        $_SERVER = [];
        $_REQUEST = [];
        //Flight::setEngine(new Engine());
        Flight::set('flight.views.path', __DIR__ . '/views');
    }

```
Well with the `new Engine()`, you have a new one for `setUp()`, but that's what we need to avoid in persistent applications. And also in tests, to run faster tests.

n0nag0n (approved):

n0nag0n:
> I see that you use it in the tests:
> 
> ```
> protected function setUp(): void
>     {
>         $_SERVER = [];
>         $_REQUEST = [];
>         Flight::init();
>         Flight::setEngine(new Engine());
>         Flight::set('flight.views.path', __DIR__ . '/views');
>     }
> ```
> 
> But when you call the `Flight::init()`, automatically set the engine with `new Engine()` if not set. I don't understand why later call `Flight::setEngine(new Engine());` and also invalidate the before `init()`.
> 
> It'll work the same with:
> 
> ```
> protected function setUp(): void
>     {
>         $_SERVER = [];
>         $_REQUEST = [];
>         //Flight::setEngine(new Engine());
>         Flight::set('flight.views.path', __DIR__ . '/views');
>     }
> ```
> 
> Well with the `new Engine()`, you have a new one for `setUp()`, but that's what we need to avoid in persistent applications. And also in tests, to run faster tests.

I tried looking into this, but when I did it, a bunch of unit tests blew up and it created some weird behavior. We may have to wait until the other PR goes into master before we can pull this one off correctly.

joanhey:
If you have your tests to work without `setUp()` the engine, the framework will work correctly with any persistent app (async).

PR #628, opened 2025-02-18
Added multipart data process for PUT, PATCH, DELETE methods

KnifeLemon:


n0nag0n:
Hey again! So with this code, you actually added quite a bit of code, and it's not unit tested so all the build steps are failing. One thought I had was that you can actually override your own request locally in your project. Something like:

```php
class CustomRequest extends flight\net\Request {
// your custom code to handle pulling the multipart data. 
}


// some bootstrap code

Flight::unregister('request');
Flight::register('request', CustomRequest::class);
```

If we include this in the framework, it's got to be unit tested and have 100% coverage and that'll take a bit with all the extra code you put in.

KnifeLemon:
I forgot the override!
I'll use it as override in the project

And I will update the code to help you with the unit test

Thank you  :)

n0nag0n:
Lemme know if you still want/need to do this after you've tested it out

PR #635, opened 2025-03-09
fix: corrected windows line endings in view tests

fadrian06:


n0nag0n (approved):

n0nag0n:
hopefully this fixes it for good cause it is super annoying when I go from my linux laptop to windows laptop and it throws errors because of \r. Thanks!

PR #638, opened 2025-03-13
Duplicated slash in groups support

### LINKED ISSUE #637: Remove multiple // in final route patterns resolving nested groups

fadrian06:
Fixes #637 

## Before
```php
Flight::group('/', static function (): void {
  Flight::route('GET /', static function (): void {});
});
```

When url is `/` it throws 404
now

## After
```php
Flight::group('/', static function (): void {
  Flight::group('/', static function (): void {
    // ...
      // ...
        // ...
          // ...
            Flight::route('GET /', static function (): void {
              echo 'works :)';
            });
  });
});
```

Works :)
image


flight/core/Dispatcher.php
```diff
@@ -437,11 +439,12 @@ protected function verifyValidClassCallable($class, $method, $resolvedClass): vo
...
-        if (
-            is_a($this->containerHandler, '\Psr\Container\ContainerInterface')
-            && $this->containerHandler->has($class)
-        ) {
-            return $this->containerHandler->get($class);
+        if (is_a($this->containerHandler, '\Psr\Container\ContainerInterface')) {
+            try {
+                return $this->containerHandler->get($class);
+            } catch (Throwable $exception) {
+                return null;
```
n0nag0n, line 446: What would happen if get() threw an error? Something should be surfaced to know what the problem was right? If say the service wasn't registered and it threw an error, but this only returned null, I feel like the end user would be confused at what was happening and it would be hard to debug.

n0nag0n (changes requested):
Outside the one try catch issue, I'm good to merge this in.

flight/core/Dispatcher.php
fadrian06: ### Without try catch
image
### With try catch
image

fadrian06:
`adhore/cli` throws a deprecation error in php 8.4 in `callable $onExit = null` must be `?callable $onExit = null`

n0nag0n (approved):

PR #644, opened 2025-04-30
test: re-run tests automatically with phpunit-watcher when a change is made

pierresh:
That is much more comfortable than restarting tests manually when a change is made

The command to run tests remains unchanged:

```bash
composer test
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,4 +20,4 @@ jobs:
                   extensions: curl, mbstring
                   tools: composer:v2
             - run: composer install
-            - run: composer test
+            - run: composer test-ci
--- a/composer.json
+++ b/composer.json
@@ -51,6 +51,7 @@
         "phpstan/phpstan": "^2.1",
         "phpunit/phpunit": "^9.6",
         "rregeer/phpunit-coverage-check": "^0.3.1",
+        "spatie/phpunit-watcher": "^1.23 || ^1.24",
         "squizlabs/php_codesniffer": "^3.11"
     },
     "config": {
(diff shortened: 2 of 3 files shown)


n0nag0n:
I love the idea, we just gotta get it working with php 7.4 and php 8. Maybe try altering the composer.json file to `"spatie/phpunit-watcher": "^1.23",` or `"spatie/phpunit-watcher": "^1.23 || ^1.24",` instead? The other thing is that I would change it to keep `composer test` the way it is, but add a `composer test-watch` or something like that cause I can see a world where watching the repo breaks the build somehow. The only other thing I thought about was this change is cool (I didn't even know it was a think) but would it make more sense in the skeleton repo with some added docs in the documentation to tell people about it?

pierresh:
> I love the idea, we just gotta get it working with php 7.4 and php 8. Maybe try altering the composer.json file to `"spatie/phpunit-watcher": "^1.23",` or `"spatie/phpunit-watcher": "^1.23 || ^1.24",` instead? The other thing is that I would change it to keep `composer test` the way it is, but add a `composer test-watch` or something like that cause I can see a world where watching the repo breaks the build somehow. The only other thing I thought about was this change is cool (I didn't even know it was a think) but would it make more sense in the skeleton repo with some added docs in the documentation to tell people about it?

I found a way to fix the conflict of required versions (1.23 / 1.24) depending on the PHP version.
 
The build failed because the github actions also relied on `composer test`. And with `phpunit-watcher`, it never stops actually, it was waiting for changes :). That is why the build failed after a while (once a timeout was reached)

I did the opposite, I did an alias dedicated to Github actions (`composer test-ci`) and kept `composer test` for `phpsunit-watcher` by default.

n0nag0n (approved):

n0nag0n:
Thanks so much!

PR #661, opened 2025-09-28
Fix(reflection): avoid deprecated setAccessible() in PHP 8.5+

arshidkv12:


n0nag0n (approved):
Thanks for catching this!

PR #666, opened 2025-11-18
Fixed an issue where `Flight::request()->base` would display only a yen symbol in Windows environments without a parent directory

riku22:
In Windows environments, when `$_SERVER['SCRIPT_NAME']` is a filename without a parent directory, such as `/index.php`, `Flight::request()->base` returns only a yen symbol. This issue will be fixed.

\* As this is my first pull request to this repository, I would appreciate it if you could point out any issues.

### DIFF
--- a/flight/net/Request.php
+++ b/flight/net/Request.php
@@ -173,6 +173,9 @@ class Request
                 $base = str_replace(['\\', ' '], ['/', '%20'], $base);
             }
             $base = dirname($base);
+            if ($base === '\\') {
+                $base = '/';
+            }
             $config = [
                 'url'        => $url,
                 'base'       => $base,


n0nag0n:
Hey there! Thanks for contributing!

I ran the workflows and it looks like all of them are failing the PHP Unit tests. The thing for you to check is when you're testing this to make sure it works locally. You can test it with `composer test` or `composer test-coverage`. This will at least checked your current version of PHP.

The thing I'm wondering about is more why you have slash characters being converted to Yen characters. Part of me is wondering if you have a weird version of PHP or something? I've never heard of that before and I've used/built things with Flight on Windows before without the issues you were talking about.

riku22:
Hello.

I apologize for the inconvenience caused by my oversight.

I have updated the pull request title and comments; I would appreciate it if you could review them.

n0nag0n (approved):
That is odd behavior for sure. It looks safe how you've implemented it and all unit tests are passing now so I think we're good to merge this in, thanks!

PR #672, opened 2025-12-27
fix  index_root  or app_root not set in .runway-config.json error

ycrao:
f0123faf-c6ac-4e20-a18f-e04389fc4ee1

The `app_root` and `index_root` values are now located in `$config['runway']` .

### DIFF
--- a/flight/commands/ControllerCommand.php
+++ b/flight/commands/ControllerCommand.php
@@ -29,7 +29,7 @@ class ControllerCommand extends AbstractBaseCommand
     public function execute(string $controller)
     {
         $io = $this->app()->io();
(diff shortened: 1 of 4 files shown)


ycrao:
ref: https://github.com/flightphp/runway/blob/master/runway#L110-L134

n0nag0n:
Hey there, thanks for doing this, although I had to do a lot more to it. I needed to still accept the old .runway-config.json file if it was still there and accept the config.php values so it took more refactoring. https://github.com/flightphp/core/pull/673

Thanks for contributing lately! Hop in the discord channel and say hi if you haven't already!

PR #675, opened 2026-01-01
Add return type declarations

arshidkv12:


n0nag0n:
I'm going to assume you wanted #676 and not this one.

PR #679, opened 2026-01-16
Ignore LiMIT when using RETURNING

Lawrence72:
Using fetchRow with RETURNING creates invalid SQL. Ignoring LIMIT will fix this issue.


n0nag0n:
Thank ya!

PR #680, opened 2026-02-19
Turns off Caching for JSON responses

lucanos:
Adds headers to json response to avoid caching of the response

### DIFF
--- a/flight/Engine.php
+++ b/flight/Engine.php
@@ -934,6 +934,9 @@ class Engine
         $this->response()
             ->status($code)
             ->header('Content-Type', 'application/json')
+            ->header('Cache-Control', 'no-store, no-cache, must-revalidate, max-age=0')
+            ->header('Pragma', 'no-cache')
+            ->header('Expires', '0')
             ->write($json);
         if ($this->response()->v2_output_buffering === true) {
             $this->response()->send();


craigrodway:
What's the rationale for having Flight set this as the default?

There are legitimate cases where developers may want to have a JSON response cached for all requests, or just for some requests and not others. I think these headers should be opt-in rather than a blanket default.

n0nag0n:
So there's been some controversy around this area of code. What you've added will explicitly not require it to be cached, but also some other behaviors make it so it wouldn't cache even with this code not added in. I used to have it added in, but then there was something else (and it escapes me what it was) but something else was causing the caching behavior so I had it removed.

I'll need to do some thinking and testing with this guy once I get some time.

fadrian06:
@lucanos what about this?
```php
$this->set('flight.cache_json', false);
```

and
```php
$this->response()->status($code)->header('Content-Type', 'application/json');

if ($this->get('flight.cache_json')) {
  $this->response()->header('Cache-Control', 'no-store, no-cache, must-revalidate, max-age=0')
    ->header('Pragma', 'no-cache')
    ->header('Expires', '0');
}

$this->response()->write($json);
```

PR #682, opened 2026-03-15
Simplify git attributes

fadrian06:
This pull request introduces several improvements to the project's configuration files, primarily to standardize and modernize the handling of code quality tools and test configuration. The main changes include adding new default configuration files for PHPStan and PHPUnit, updating `.gitattributes` to better manage exported files, and enhancing the Composer scripts to ensure necessary config files are available after installation.

**Configuration management improvements:**

* Added `phpstan.dist.neon` as the new default PHPStan configuration file, specifying analysis level, included paths, and exclusions.
* Added `phpunit.xml.dist` as the new default PHPUnit configuration file, defining test suites, coverage options, and strictness settings.
* Updated Composer scripts in `composer.json` to automatically copy `phpstan.dist.neon` to `phpstan.neon` and `phpunit.xml.dist` to `phpunit.xml` if they do not exist, ensuring consistent local configuration for developers.

**Export and distribution adjustments:**

* Updated `.gitattributes` to exclude new configuration files (`phpstan.dist.neon`, `phpunit.xml.dist`, etc.) and other files from package exports, while removing obsolete or renamed entries.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+/.gemini export-ignore
 /.github export-ignore
 /tests export-ignore
 /.editorconfig export-ignore
@@ -5,5 +6,8 @@
 /.gitignore export-ignore
 /phpcs.xml export-ignore
 /phpstan.neon export-ignore
+/CONTRIBUTING.md export-ignore
+/index.php export-ignore
 /phpstan-baseline.neon export-ignore
 /phpunit.xml export-ignore
+/phpunit-watcher.yml export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,10 +4,10 @@
 /.editorconfig export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore
-/phpcs.xml export-ignore
-/phpstan.neon export-ignore
 /CONTRIBUTING.md export-ignore
 /index.php export-ignore
+/phpcs.xml.dist export-ignore
 /phpstan-baseline.neon export-ignore
-/phpunit.xml export-ignore
+/phpstan.dist.neon export-ignore
 /phpunit-watcher.yml export-ignore
+/phpunit.xml.dist export-ignore
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ coverage/
 *.sublime*
 clover.xml
 phpcs.xml
+phpstan.neon
+phpunit.xml
 .runway-config.json
 .runway-creds.json
 .DS_Store
--- a/composer.json
+++ b/composer.json
@@ -82,11 +82,13 @@
             "rm server.pid",
             "echo \"Performance Tests Completed.\""
         ],
-        "lint": "phpstan --no-progress --memory-limit=256M -cphpstan.neon",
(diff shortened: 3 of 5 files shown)


n0nag0n (changes requested):
I'm not totally sure why you like setting up .dist files and then copying them over as the real file and ignoring it. I understand that you CAN do it, not sure why it needs to be done or what real benefit it has besides just adding some more complexity to things. Maybe I just need to ask @copilot and he'll explain it to me haha

fadrian06:
ok, I'll tell you my real usage of .dist files, not theory unusable...

> I'm working on a project with phpstan.dist.neon file, that means that all contributors must follow that phpstan restrictions, mentioned in the dist file (distributed file)...

> but, because project has a dist file, I can copy it to a local phpstan.neon file and add more restrictions, more rules, to have a cleaner code than other contributors por example, that's the real usage.

fadrian06:
the most useful use case is for the phpunit.xml.dist

> in phpunit configuration file, you can set environment variables for testing, instead of using .env files, so in the .dist file, you can set placeholder or common values for database connection, or configuration or something else.

> but the advantage of that is that now I can create a local phpunit.xml (not dist) to customize that environment variables or customize phpunit a little different of the others contributors.

n0nag0n:
ok, I actually can get behind that! The one change I would ask then is if you could change the name to phpstan.neon.dist instead. Looks like that's more standard around projects (similar to how you have it for phpunit.xml.dist). Then I can merge this guy.

fadrian06:
> ok, I actually can get behind that! The one change I would ask then is if you could change the name to phpstan.neon.dist instead. Looks like that's more standard around projects (similar to how you have it for phpunit.xml.dist). Then I can merge this guy.

<a href="https://phpstan.org/config-reference#config-file">
If you do not provide a config file explicitly, PHPStan will look for files named <code>phpstan.neon</code>, <code>phpstan.neon.dist</code>, or <code>phpstan.dist.neon</code> in the current directory.

The resolution priority is as such:
<ol>
<li>If config file is provided as a command line option, it will be used.
<li>Otherwise, if phpstan.neon exists in the current working directory, it will be used.
<li>Otherwise, if phpstan.neon.dist exists in the current working directory, it will be used.
<li>Otherwise, if phpstan.dist.neon exists in the current working directory, it will be used.
<li>If none of the above is true, no config will be used.
</ol>
The usual practice is to have phpstan.neon.dist or phpstan.dist.neon under version control, and allow the user to override certain settings in their environment (on their own computer or on a continuous integration server) by creating phpstan.neon that’s present in .gitignore file.
</a>

fadrian06:
> ok, I actually can get behind that! The one change I would ask then is if you could change the name to phpstan.neon.dist instead. Looks like that's more standard around projects (similar to how you have it for phpunit.xml.dist). Then I can merge this guy.

I use phpstan.dist.neon because my code editor highlight neon syntax
image

If I use phpstan.neon.dist, my editor can't infer that file as a neon file, if I force that, SublimeText will highlight phpunit.xml.dist as a neon file as well instead of a xml file
image

n0nag0n:
as long as stuff doesn't blow up..... :sweat:

PR #694, opened 2026-04-23
fix wildcard routes being skipped by fast path method lookup closes #693

### LINKED ISSUE #693: Method-specific route matching in v3.17.1 affects the order
**BUG description**
A change occurred between `v3.17.0` and `v3.17.1` introduced a bug that is affecting the order in which Routes are matched.
This is likely to be caused by the new feature #658 that modifies the routing algorithm to check method-specific routes first.

**To Reproduce**
Sample code to reproduce the bug:

```php
Flight::route("/@par/[^\/]+/*", function(string $par){
    echo "Passthrough (Par = $par) <br>";
    return true;
});

Flight::route("GET /[^\/]+/target", function(){
    echo "Target";
});

Flight::start();
```

**Expected behavior**
Expected result (working until `v.3.17.0`):
```
Passthrough (Par = foobar)
Target
```

Starting from `v.3.17.1` the first route defined does NOT get matched first and the output is:
```
Target
```

**Additional context**
That regex is structured in that way in order to capture that first parameter in the URI, make some validation/adjustments and continue to the actual route, no matter what method is used.

I am fully aware that probably middleware would be more appropriate for flight-specific projects, but this is a stripped-down example from a very complex project that is currently relying on the assumption that ***Routes are matched in the order they are defined***.

If keeping the original behaviour while preserving the new method-specific route matches is impossible, please consider adding a configuration option to turn off the new feature. This was definitely a breaking change for some projects.

KnifeLemon:
Fixes the route registration order regression introduced in v3.17.1 by the method-specific fast path added in #658.

closes #693

## Fix

Store routes in `routesByMethod` using their **global index** in `$routes` as the array key. When building the fast path candidates, merge the two buckets and `ksort` by key to restore registration order.

```php
// map() — use global route index as key
$this->routesByMethod[$method][count($this->routes) - 1] = $route;

// route() — merge and sort by registration order
$candidates = ($this->routesByMethod[$requestMethod] ?? []) + ($this->routesByMethod['*'] ?? []);
ksort($candidates);
```

Now the merged candidates are always in registration order regardless of method, and the fast path performance benefit (only URL-matching against eligible routes) is preserved.

## Testing

Added regression test `testWildcardPassthroughRouteBeforeSpecificGetRoute` in `RouterTest`.

```
> phpunit --filter RouterTest
PHPUnit 9.6.34 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.29
Configuration: C:\FlightPHP\core\phpunit.xml.dist
Random Seed:   1776965395

................................................................. 65 / 66 ( 98%)
.                                                                 66 / 66 (100%)

Time: 00:00.031, Memory: 8.00 MB

OK (66 tests, 87 assertions)
```

### DIFF
--- a/flight/net/Router.php
+++ b/flight/net/Router.php
@@ -152,7 +152,7 @@ class Router
             if (!isset($this->routesByMethod[$method])) {
                 $this->routesByMethod[$method] = [];
             }
-            $this->routesByMethod[$method][] = $route;
+            $this->routesByMethod[$method][count($this->routes) - 1] = $route;
         }
 
         return $route;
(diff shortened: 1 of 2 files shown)


n0nag0n (approved):
If everything is still passing, should be good to go! Thank you!

PR #702, opened 2026-06-22
Update workflow to support multiple OS environments

fadrian06:
This pull request updates the test workflow configuration to expand the operating systems used in the test matrix. Now, tests will run on Ubuntu, Windows, and macOS environments for each PHP version.

Test matrix improvements:

* Added an `os` matrix to the workflow, so tests are now run on `ubuntu-latest`, `windows-latest`, and `macos-latest` for each PHP version. Updated the `runs-on` field to use the matrix value. (.github/workflows/test.yml)


n0nag0n:
Thanks for fixing that so the tests work for windows now too. Hopefully this helps you out as well!

